PR #20, opened 2022-03-12
Flockpanel TGUI

### LINKED ISSUE #10: Swap from CHUI to TGUI
### Labels

[FEATURE]

### Description

CHUI old, TGUI new and shiny

### Steps to reproduce

1. Open spessgame
2. see CHUI windows everywhere
3. try not to cry
4. cry a lot

### Expected Behavior

TGUI

_No response_

TobleroneSwordfish:
[QOL]
## About the PR 
Converts the flockpanel UI from old nasty CHUI to new shiny TGUI. 
Fixes #10 
drones
partitions
structures
enemies
Task icon list: 
image

Known issues:
- Structures don't report type specific information because that's likely to change a decent amount

Currently the UI updates by iterating the entire flock in `ui_data` every second, which is definitely not going to cut it performance wise. I'm thinking maybe we maintain UI data caches on the backend and update them dynamically, unless it's possible to get TGUI to take partial data updates? Basically I need advice from people who know how to optimise TGUI. (unless this is actually fine and I overestimate TGUI performance impact)
## Why's this needed? 
tgui good


stonepillars:
!merge_upstream

code/datums/abilities/flock/flockmind.dm
```diff
@@ -299,13 +299,17 @@
...
-		return 1
-	panel.Subscribe(user)
+	F.flock.flockpanel = tgui_process.try_update_ui(F, F, F.flock.flockpanel)
+	if (!F.flock.flockpanel)
+		F.flock.flockpanel = new(F, F, "FlockPanel")
+		F.flock.flockpanel.open()
+	// var/chui/window/flockpanel/panel = F.flock.panel
+	// if(isnull(user) || isnull(F) || isnull(panel))
+		// return 1
+	// panel.Subscribe(user)
```
stonepillars, line 312:
```suggestion
```
Same with all of the other commented out CHUI, just remove them, don't need the comments since we have the blame/diff.

TobleroneSwordfish: good point, I'll strip out all the CHUI code

amylizzle:
Regarding updates when drones task change, doesn't TGUI update every second anyway? Do we really need <1 second resolution on task updates?

TobleroneSwordfish:
> Regarding updates when drones task change, doesn't TGUI update every second anyway? Do we really need <1 second resolution on task updates?

Currently the 1 second updates aren't working with drone tasks, although now I think about it I'm not sure why they don't.

amylizzle:
also link this to #10

TobleroneSwordfish:
Alright, fixed the update issues because I'm an idiot and didn't add one of the TGUI boilerplate procs.

amylizzle (changes requested):

code/mob/living/intangible/flock/flockmind.dm
```diff
@@ -11,6 +11,56 @@
...
+				if(alert(src, "This will destroy the flocktrace. Are you ABSOLUTELY SURE you want to do this?", "Confirmation", "Yes", "No") == "Yes")
+					// if they're in a drone, kick them out
+					var/mob/living/critter/flock/drone/host = T.loc
+					if(istype(host))
+						host.release_control()
+					// DELETE
+					flock_speak(null, "Partition [T.real_name] has been reintegrated into flock background processes.", src.flock)
+					boutput(T, "<span class='flocksay'><b>\[SYSTEM: Your higher cognition has been forcibly reintegrated into the collective will of the flock.\]</b></span>")
+					T.death()
+
```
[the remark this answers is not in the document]
TobleroneSwordfish: trace.death() calls qdel in this case, so it's not required

amylizzle (approved):

tgui/packages/tgui/interfaces/FlockPanel.js
```diff
@@ -0,0 +1,324 @@
...
+        {tabIndex === 1
+        && (
+          <Box>
+            <Dropdown
+              options={["name", "health", "resources", "area"]}
+              selected="resources"
+              onSelected={(value) => setSortBy(value)}
+            />
+            <FlockDrones drones={drones} sortBy={sortBy} />
+          </Box>
```
stonepillars, line 316: A little weird that this isn't driven by local state on the FlockDrones component and is instead passed here, but I don't know much about tgui.

TobleroneSwordfish: Yeah, it's passed in so I don't have to put the dropdown into FlockDones mostly just so it matches the other Flock components, no idea which is better practice, though it should probably at least be called "sortDrones" or something if it is left outside.

stonepillars: OH WELL WHAT COULD POSSIBLY GO WRONG

PR #30, opened 2022-03-13
Apply pali's set_area fix

TobleroneSwordfish:
[BUG]
## About the PR 
Fixes the annoying "set_area is undefined" errors on startup that can cause you to have to restart 6 times to be able to actually test something.

## Why's this needed? 
my sanity

### DIFF
--- a/code/datums/titlecard.dm
+++ b/code/datums/titlecard.dm
@@ -59,7 +59,8 @@
 	if (last_pregame_html == pregameHTML)
 		for(var/client/C)
 			if(istype(C.mob, /mob/new_player))
-				C << output(list2params(list(id, text)), "pregameBrowser:set_area")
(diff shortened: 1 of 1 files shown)


TobleroneSwordfish:
oh goddamnit why does that fail

stonepillars:
I'm assuming this is a rebase of a commit from goonstation mainline, you can just merge goonstation mainline as a PR. I can do it later if you don't want to screw with that.

TobleroneSwordfish:
> I'm assuming this is a rebase of a commit from goonstation mainline, you can just merge goonstation mainline as a PR. I can do it later if you don't want to screw with that.

Huh, I couldn't see how to do that so I just used git-cherrypick. Go ahead if you know a cleaner way.

PR #33, opened 2022-03-13
Flock Compute Base

amylizzle:
## About the PR 
Establishes the base of the compute resource, adding a couple of utility functions and a UI element to the flockmind/flocktrace ability holder.

Also added the check that ensures flocktraces die when there is insufficient compute power.

Collectors also provisionally generate compute based on the number of tiles they were connected to. This should probably be changed, but makes a handy test case until we decide how it should change.

Sentinels now check to see if there is 20 available compute, and are active if this is the case. Previously this was a check on power.

Notably the code for tilegroups seems to have been exclusively used for determining power generation/consumption in a localised way. I've taken the power part out but left the tile grouping code in. Maybe we can use it for something in future or remove it at a later date.

## Why's this needed? 
Setting up for the midgame

### DIFF
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -3,8 +3,29 @@
 ///////////////////////
 
 /datum/abilityHolder/flockmind
+	tabName = "Flockmind"
+	usesPoints = 1
+	points = 0 //total compute - used compute
+	var/tC = 0 //total compute
+	regenRate = 0
 	topBarRendered = 1
 	rendered = 1
+	notEnoughPointsMessage = "<span class='alert'>Insufficient available compute resources.</span>"
+
+/datum/abilityHolder/flockmind/proc/updateCompute()
+	var/mob/living/intangible/flock/flockmind/F = owner
+	if(!F?.flock)
+		return //someone made a flockmind without a flock, or gave this ability holder to something else.
+	src.tC = F.flock.total_compute()
+	var/uC = F.flock.used_compute()
+	src.points = src.tC - uC
+
+/datum/abilityHolder/flockmind/onAbilityStat()
+	..()
+	.= list()
+	.["Compute:"] = "[round(src.points)]/[round(src.tC)]"
+	//.["Total Compute:"] = round(F.flock?.total_compute())
+	return
 
 /atom/movable/screen/ability/topBar/flockmind
 	tens_offset_x = 19
@@ -120,6 +141,9 @@
 
 /datum/targetable/flockmindAbility/partitionMind/cast(atom/target)
 	if(..())
+		return 1
+	if(!holder.pointCheck(100))
+		boutput(holder,"<span class='alert'>You need more compute to partition yourself</span>")
 		return 1
 	var/mob/living/intangible/flock/flockmind/F = holder.owner
 	if(F)
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -6,7 +6,7 @@
 	tabName = "Flockmind"
 	usesPoints = 1
 	points = 0 //total compute - used compute
-	var/tC = 0 //total compute
+	var/totalCompute = 0
 	regenRate = 0
 	topBarRendered = 1
 	rendered = 1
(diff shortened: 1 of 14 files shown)


code/datums/abilities/flock/flockmind.dm
```diff
@@ -3,8 +3,29 @@
 ///////////////////////
 
 /datum/abilityHolder/flockmind
+	tabName = "Flockmind"
+	usesPoints = 1
+	points = 0 //total compute - used compute
+	var/tC = 0 //total compute
```
TobleroneSwordfish, line 9: Just call it total_compute, I know flock code is full of 1 or 2 letter vars but we can be better than that

code/datums/flock/flock.dm
```diff
@@ -151,6 +152,43 @@
 		res += F.resources
 	return res
 
+/datum/flock/proc/total_compute()
+	var/res = 0
```
TobleroneSwordfish, line 156: Call it result or just use .

code/datums/flock/flock.dm
```diff
@@ -151,6 +152,43 @@
...
+	var/tmp = 0
```
TobleroneSwordfish, line 157: What is tmp?

code/datums/flock/flock.dm
```diff
@@ -261,15 +303,32 @@
...
+	if(isflock(D))
 		src.units -= D
 
 		if(D:real_name && busy_tiles[D:real_name])
 			src.busy_tiles[D:real_name] = null
+		var/datum/abilityHolder/flockmind/aH = src.flockmind.abilityHolder
+		aH.updateCompute()
+// STRUCTURES
+
+/datum/flock/proc/registerStructure(var/atom/movable/D)
```
TobleroneSwordfish, line 321: D?

code/datums/flock/flock.dm
```diff
@@ -151,6 +152,43 @@
...
+		tmp = F.compute_provided()
+		if(tmp<0)
+			res += abs(tmp)
+
+	for(var/obj/flock_structure/S in src.structures)
+		tmp = S.compute_provided()
+		if(tmp<0)
+			res += abs(tmp)
+
+	//not strictly necissary, but maybe future traces can provide compute in some way or cost more when doing stuff?
```
TobleroneSwordfish, line 182:
```suggestion
	//not strictly necessary, but maybe future traces can provide compute in some way or cost more when doing stuff?
```

code/datums/flock/flocktilegroup.dm
```diff
@@ -24,7 +22,7 @@
 	..() //linter machine go ANGRY
 
 /datum/flock_tile_group/proc/process()
```
TobleroneSwordfish, line 24: If it just calls the parent then it doesn't need to exist

code/datums/abilities/flock/flockmind.dm
```diff
@@ -3,8 +3,29 @@
...
 	topBarRendered = 1
 	rendered = 1
+	notEnoughPointsMessage = "<span class='alert'>Insufficient available compute resources.</span>"
+
+/datum/abilityHolder/flockmind/proc/updateCompute()
+	var/mob/living/intangible/flock/flockmind/F = owner
+	if(!F?.flock)
+		return //someone made a flockmind without a flock, or gave this ability holder to something else.
+	src.tC = F.flock.total_compute()
+	var/uC = F.flock.used_compute()
```
TobleroneSwordfish, line 20: same here

TobleroneSwordfish (changes requested):
Have a review that is mostly just me complaining about variable names

code/datums/flock/flocktilegroup.dm
stonepillars: More specifically, it doesn't need to be re-defined for the subtype.

amylizzle: weirdly it seems to need to exist, there's a call on all datums:process() deep in the code, and I guess /datum doesn't implement it. So it isn't an override, and should just return 0, not call the parent.

TobleroneSwordfish: I hate it

amylizzle: me too. We can probably remove this entire thing if we're ditching the power stuff tbh.

I'm hesitant to remove it unless we're sure though.

code/datums/flock/flock.dm
```diff
@@ -146,10 +147,46 @@
...
+		if(comp_provided<0)
+			. += abs(comp_provided)
+
+	for(var/obj/flock_structure/S in src.structures)
+		comp_provided = S.compute_provided()
+		if(comp_provided<0)
+			. += abs(comp_provided)
+
+	//not strictly necessary, but maybe future traces can provide compute in some way or cost more when doing stuff?
+	for(var/mob/living/intangible/flock/trace/T in src.traces)
```
ZeWaka, line 183: as anything

PR #34, opened 2022-03-14
Added a drone resource display and moved some HUD elements around to make room

### LINKED ISSUE #31: Drones need a UI element showing resources
### Labels

[FEATURE]

### Description

As a drone, you only see how many resources you have when you finish consuming an item

### Steps to reproduce

1. Be a drone
2. Have no idea how many resources you have

### Expected Behavior

Maybe something akin to vampire's blood counter UI element - you can do this through abilityHolder

_No response_

TobleroneSwordfish:
[QOL]
## About the PR 
Fixes #31 and moves the drone eject button to the bottom left to match the AI shell eject button.

## Why's this needed? 
Knowing how many resources you have as a drone is good

### DIFF
--- a/code/datums/abilities/flockdrone.dm
+++ b/code/datums/abilities/flockdrone.dm
@@ -1,0 +1,24 @@
+/datum/abilityHolder/critter/flockdrone
+	//this is jank, but we want to keep the resources stored on the drone not the abilityholder so we do this
+	var/mob/living/critter/flock/drone/flockdrone
+	var/resources_last = -1
+
+	New()
+		. = ..()
+		if (istype(owner,/mob/living/critter/flock/drone))
+			flockdrone = owner
+
+	onAbilityStat()
+		..()
+		if (!flockdrone)
+			return
+		.= list()
+		.["Resources:"] = flockdrone.resources
+
+	onLife()
+		. = ..()
+		if (!flockdrone)
+			return
+		if (resources_last != flockdrone.resources)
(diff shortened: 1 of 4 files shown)


stonepillars:
When you go above 100 resources, it breaks onto another line.
image

code/datums/abilities/flockdrone.dm
```diff
@@ -0,0 +1,24 @@
+/datum/abilityHolder/critter/flockdrone
+	//this is jank, but we want to keep the resources stored on the drone not the abilityholder so we do this
+	var/mob/living/critter/flock/drone/flockdrone
+	var/resources_last = -1
+
+	New()
+		. = ..()
+		if (istype(owner,/mob/living/critter/flock/drone))
```
ZeWaka, line 8: this should throw a runtime or something if it's not a flock

TobleroneSwordfish:
> When you go above 100 resources, it breaks onto another line.

Yep, that's just how onAbilityStat works I think. It just so happens that all other antag point types have shorter names (blood, energy etc.)
Honestly it might make more sense if it was called "matter" anyway to distinguish it from compute.

ZeWaka:
yeah status panel sucks ass

amylizzle (approved):

stonepillars:
> Honestly it might make more sense if it was called "matter" anyway to distinguish it from compute.

eh, flockmind is the one who has to deal with compute counts.

PR #49, opened 2022-03-15
Fix designating as enemy things that shouldn't be designated using the Designate Enemy ability.

### LINKED ISSUE #43: Flockdrones and Flockbits can be designated as enemies
### Labels

[BUG]

### Description

You can mark Flockdrones and Flockbits as enemies.

### Steps to reproduce

1. Be Flockmind
2. Select Designate Enemy ability
3. Select a Flockdrone or Flockbit

### Expected Behavior

They shouldn't be able to be designated as enemies.

_No response_

FlameArrow57:
[BUG]

## About the PR 
This PR prevents designating things that shouldn't be designated as an enemy with the Designate Enemy ability.

## Why's this needed? 
Bug fix, fixes #43.

### DIFF
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -99,16 +99,19 @@
 /datum/targetable/flockmindAbility/designateEnemy/cast(atom/target)
 	if(..())
 		return 1
+
 	var/mob/M = target
 	var/mob/living/intangible/flock/flockmind/F = holder.owner
-	if(isliving(M) && !isflock(M))
-		if(F)
-			var/datum/flock/flock = F.flock
-			if(flock)
-				flock.updateEnemy(M)
-	else
-		boutput(holder.owner, "<span class='alert'>That isn't a valid target.</span>")
+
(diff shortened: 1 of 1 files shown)


FlameArrow57:
I'd like to add other things that shouldn't be designated in the same PR.

code/datums/abilities/flock/flockmind.dm
```diff
@@ -99,9 +99,9 @@
 /datum/targetable/flockmindAbility/designateEnemy/cast(atom/target)
 	if(..())
 		return 1
-	var/mob/living/M = target
+	var/mob/M = target
 	var/mob/living/intangible/flock/flockmind/F = holder.owner
-	if(isliving(M))
+	if(isliving(M) && !isflock(M))
```
amylizzle, line 104: maybe also !isintangible(M) so we don't accidentally mark the AI eye as a target.

stonepillars: This, but also convert this if into a guard (i.e. early return) instead of it being a block containing all the logic

stonepillars (approved):

PR #54, opened 2022-03-16
Restrict Flockmind rift placement

### LINKED ISSUE #52: Spawning a flock rift in space
### Labels

[BUG]

### Description

You are able to spawn a flock rift in space and it flings your drones off to wander the edges of the station z for ever

### Steps to reproduce

1. Be flockmind
2. Spawn your initial flock rift in space
3. Drones are flung off at lightspeed to the edges of the map and runtime when they land
4. Drones are now too far away to rally to the station and so just wander around for ever unless manually brought back

### Expected Behavior

Probably best if flock have to start on a station tile

_No response_

FlameArrow57:
## About the PR 
This PR restricts placement of the initial Flockmind rift. The conditions that the initial blob nucleus placement has are used, along with solar panel and mining magnet areas to prevent rift spawned things from being flung out into space. There may be a few cases on maps where stuff could still be flung out into space, but this should help prevent most of it I think.

## Why's this needed? 
Flockmind rift placement isn't restricted right now.

Fixes #52.

### DIFF
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -57,10 +57,38 @@
 
 /datum/targetable/flockmindAbility/spawnEgg/cast(atom/target)
 	if(..())
-		return 1
+		return
 	var/mob/living/intangible/flock/flockmind/F = holder.owner
-	if(F)
-		F.spawnEgg()
+
+	var/turf/T = get_turf(F)
+
+	if (istype(T, /turf/space/) || istype(T.loc, /area/station/solar) || istype(T.loc, /area/station/mining/magnet))
+		boutput(F, "<span class='alert'>Space and exposed areas are unsuitable for rift placement!</span>")
+		return
+
+	if (!isadmin(F))
+		if(IS_ARRIVALS(T.loc))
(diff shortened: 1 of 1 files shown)


code/datums/abilities/flock/flockmind.dm
```diff
@@ -57,10 +57,38 @@
 
 /datum/targetable/flockmindAbility/spawnEgg/cast(atom/target)
 	if(..())
-		return 1
+		return
```
TobleroneSwordfish, line 60: I think returning 1 (or TRUE) indicates a fail state for abilities

FlameArrow57: Yeah, the parent function is this I think
image

FlameArrow57: > I think returning 1 (or TRUE) indicates a fail state for abilities

I just removed it because there was no cooldown with the ability, but I can add it back if that's better.

stonepillars: Always use TRUE

amylizzle (approved):

PR #57, opened 2022-03-16
Prevent creating tealprints on tiles that have something with density

FlameArrow57:
## About the PR 
This PR prevents creating tealprints on tiles that have something with density, excluding Flockdrones since they can move through fully constructed tealprint structures.

It also adds a different message for if there's a tealprint scheduled on a tile versus a finished tealprint structure when trying to schedule a tealprint.

## Why's this needed? 
It doesn't make sense that you can schedule a tealprint on a tile containing something with density.

### DIFF
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -351,7 +351,7 @@
 		return 1
 	for (var/atom/O in T.contents)
 		if (O.density && !isflock(O))
-			boutput(holder.owner, "<span class='alert'>A tealprint cannot be scheduled with [O] on the same tile!</span>")
+			boutput(holder.owner, "<span class='alert'>That tile has something that blocks tealprint creation!</span>")
 			return 1
 	//todo: replace with FANCY tgui/chui window with WHEELS and ICONS and stuff!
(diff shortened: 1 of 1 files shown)


code/datums/abilities/flock/flockmind.dm
```diff
@@ -343,9 +343,16 @@
...
 		return 1
+	if(locate(/obj/flock_structure/ghost) in T)
+		boutput(holder.owner, "<span class='alert'>A tealprint has already been scheduled here!</span>")
+		return 1
 	if(locate(/obj/flock_structure) in T)
 		boutput(holder.owner, "<span class='alert'>There is already a flock structure on this flocktile!</span>")
 		return 1
+	for (var/atom/O in T.contents)
+		if (O.density && !isflock(O))
+			boutput(holder.owner, "<span class='alert'>A tealprint cannot be scheduled with [O] on the same tile!</span>")
```
stonepillars, line 354:
```suggestion
			boutput(holder.owner, "<span class='alert'>A tealprint cannot be scheduled with \a [O] on the same tile!</span>")
```

FlameArrow57: > 

\a or \the? Some objects already have "the" in their name I think, so maybe \the would be a little better for consistency?

stonepillars:
Change it to 'That tile has something blocking it!'. Less ambiguous

PR #61, opened 2022-03-17
Prevent Flockdrones and Flockbits from being moved by conveyors and drivers

FlameArrow57:
[FEATURE]

## About the PR 
This PR just makes it so that Flockdrones and Flockbits can't be moved by conveyor belts and mass driver type pushers.

## Why's this needed? 
It doesn't make sense considering Flockdrones and Flockbits are floating. 

It also wouldn't be good to lose them if they move onto these objects considering how valuable they can be.

### DIFF
--- a/code/WorkInProgress/recycling/conveyor.dm
+++ b/code/WorkInProgress/recycling/conveyor.dm
@@ -106,6 +106,8 @@
 		return
 	if(istype(A, /obj/critter) && A:flying)		//They are flying above it, ok.
 		return
+	if(isflock(A))
+		return
 	var/movedir = dir	// base movement dir
 	if(divert && dir == divdir)	// update if diverter present
 		movedir = divert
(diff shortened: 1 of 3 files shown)


amylizzle:
I don't entirely agree with this for a couple reasons:
1) It's hilarious watching the drones get bounced around the delivery system
2) It means that popping the rift in the core of disposals/cargo-belt-hell is always a great strategy because nobody goes there and there's no risk.

stonepillars:
Keep this open and organic because it may end up being merged if it becomes a problem during testing.

stonepillars:
IT seems... this already is the behavior. I think. I don't remember ever seeing flockdrones being moved by conveyors. Wack.

FlameArrow57:
It's because of the new AI I think, if you possess a drone and move over a conveyor you'll still be moved

amylizzle:
Maybe we just merge this for consistency. Having the drones behave differently on conveyors if someone is possessing them is bad.

stonepillars:
damn birds always flying n shit

PR #63, opened 2022-03-17
Unlockable Flock Structures! Also Endgame!

amylizzle:
## About the PR 
Changes the structures ability to support unlockable structures. You can now create a /datum/unlockable_flock_structure to add a structure to the list, announce it to the flock, and make it buildable on a condition defined in the the datum.
Flock structures now have a resourcecost property which tells the tealprint how much resource is needed to build the structure.
Tealprints now place the icon correctly, this is basically only useful for the relay but it might be for other stuff later.

Also the relay now unlocks at 1000 compute - WE HAVE AN ENDGAME OMG

## Why's this needed? 
Next step in getting towards making this shit playable.

```changelog
(u)Amylizzle
(*)You can now construct the relay as the Flock! That means you can actually win! Yay!
```


code/datums/flock/flock.dm
```diff
@@ -16,6 +16,7 @@
 	var/list/annotation_viewers = list()
 	var/list/annotations = list() // key is atom ref, value is image
 	var/list/obj/flock_structure/structures = list()
+	var/list/obj/flock_structure/unlockedStructures = list(/obj/flock_structure/collector, /obj/flock_structure/sentinel) //list of flock_structure type paths
```
stonepillars, line 19:
```suggestion
	var/list/unlockedStructures = list(/obj/flock_structure/collector, /obj/flock_structure/sentinel) //list of flock_structure type paths
```
type paths are not object instances

code/datums/abilities/flock/flockmind.dm
```diff
@@ -353,21 +351,30 @@
...
+
 	for (var/atom/O in T.contents)
 		if (O.density && !isflock(O))
 			boutput(holder.owner, "<span class='alert'>That tile has something that blocks tealprint creation!</span>")
 			return 1
+
+	var/list/friendlyNames = list()
+	var/mob/living/intangible/flock/flockmind/F = holder.owner
+	for(var/obj/flock_structure/S as anything in F.flock.unlockedStructures)
+		friendlyNames += initial(S.flock_id) //flock_id is the friendly name for the structure
```
stonepillars, line 363: This is a little weird. Maybe datumize?

code/datums/abilities/flock/flockmind.dm
```diff
@@ -353,21 +351,30 @@
...
 		return TRUE
-	switch(structurewanted)
-		if("Collector")
-			structurewantedtype = /obj/flock_structure/collector
-			resourcecost = 200
-		if("Sentinel")
-			structurewantedtype = /obj/flock_structure/sentinel
-			resourcecost = 300
+	var/obj/flock_structure/structurewantedtype = null
+	for(var/obj/flock_structure/S as anything in F.flock.unlockedStructures)
```
stonepillars, line 372:
```suggestion
	for(var/S in F.flock.unlockedStructures)
```
`S` is also a weird name, maybe change to `structure_type` since literal type paths are very unclear in DM.

code/datums/flock/flock.dm
```diff
@@ -444,19 +455,13 @@
...
-		// 	T.off()
-		// else if(validNeighbors >= 4 && !T.generating)
-		// 	T.on()
 
 	if(floors_no_longer_existing.len > 0)
 		src.all_owned_tiles -= floors_no_longer_existing
 
+	if(src.total_compute() >= 1000)
+		src.unlockStructure(/obj/flock_structure/relay)
+
```
stonepillars, line 464: This will become a giant if chain very quickly

stonepillars (changes requested):

code/datums/abilities/flock/flockmind.dm
```diff
@@ -390,21 +388,33 @@
...
 	if (!structurewanted)
-		return TRUE
-	switch(structurewanted)
-		if("Collector")
-			structurewantedtype = /obj/flock_structure/collector
-			resourcecost = 200
-		if("Sentinel")
-			structurewantedtype = /obj/flock_structure/sentinel
-			resourcecost = 300
+		return 1
```
stonepillars, line 409:
```suggestion
		return TRUE
```

code/datums/abilities/flock/flockmind.dm
```diff
@@ -390,21 +388,33 @@
...
+
 	for (var/atom/O in T.contents)
 		if (O.density && !isflock(O))
 			boutput(holder.owner, "<span class='alert'>That tile has something that blocks tealprint creation!</span>")
 			return 1
+
+	var/list/friendlyNames = list()
+	var/mob/living/intangible/flock/flockmind/F = holder.owner
+	for(var/datum/unlockable_flock_structure/ufs as anything in F.flock.unlockableStructures)
+		if(ufs.check_unlocked())
```
stonepillars, line 400: ufs has `ufs.unlocked` doesn't it?

code/datums/abilities/flock/flockmind.dm
```diff
@@ -390,21 +388,33 @@
...
+			var/obj/flock_structure/sT = ufs.structType //this is a gross hack, but needed for resolving flock_id
+			friendlyNames += initial(sT.flock_id) //flock_id is the friendly name for the structure
```
stonepillars, line 402: Consider adding the 'friendly name' to the datum as a var. It can either get the initial flock_id from the structType it was initialized with or flock_id could be moved completely to the datum, though this would be troublesome for the inspect messages.

code/datums/flock/flock.dm
```diff
@@ -447,24 +466,25 @@
...
 	src.claimTurf(flock_convert_turf(T))
 	playsound(T, "sound/items/Deconstruct.ogg", 70, 1)
 
+///Unlock an achievement (string) if it isn't already unlocked
+/datum/flock/proc/achieve(var/str)
+	src.achievements |= str
+
+///Unlock an achievement (string) if it isn't already unlocked
+/datum/flock/proc/hasAchieved(var/str)
+	return (str in src.achievements)
```
stonepillars, line 487: Probably should convert into macros, though probably should ask in #imcoder about that

code/datums/abilities/flock/flockmind.dm
amylizzle: unlocked is updated every processing cycle, check_unlocked() gets a current value. If I just get unlocked, there's a ~3 second window where you could potentially build a structure you shouldn't be able to. For the cost of a single function call, it's worth the check.

code/datums/flock/flock.dm
amylizzle: how so? achievements are going to be flock specific so if it was macro'd we'd just have to do hasAchieved(flock, "thing") instead of flock.hasAchieved("thing")

stonepillars: I'm not to sure on this one really. This is probably fine actually.

stonepillars (approved):

PR #82, opened 2022-03-19
Fix not being able to designate space tiles

FlameArrow57:
[BUG]

## About the PR 
This PR fixes not being able to priority designate space tiles as Flockmind.

## Why's this needed? 
Bug fix


stonepillars:
er, what actually weird disjoint between desc and title whoops

PR #83, opened 2022-03-19
Fix depositing resources to a tealprint when you have none

### LINKED ISSUE #14: Controlled drones can deposit materials while they have none
### Labels

[BUG]

### Description

^

### Steps to reproduce

1. Place a tealprint
2. Control a drone
3. Run out of resources
4. Try to add resources to the tealprint with your nanite spray
5. Get the action bar and message "You deposit materials to the tealprint" even though nothing actually happens

### Expected Behavior

Fail state

This shouldn't be able to happen with AI controlled drones since the AI task to contribute materials checks if the drone has materials before starting.

FlameArrow57:
[BUG]

## About the PR 
This PR prevents you from attempting to deposit resources to a tealprint when controlling a Flockdrone and you have none.

## Why's this needed? 
Bug fix, fixes #14.

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -789,7 +789,10 @@
 					else
 						actions.start(new/datum/action/bar/flock_repair(F), user)
 			if(/obj/flock_structure/ghost)
-				actions.start(new /datum/action/bar/flock_deposit(target), user)
+				if (user.resources == 0)
+					boutput(user, "<span class='alert'>No resources available for construction.</span>")
+				else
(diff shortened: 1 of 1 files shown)


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -789,7 +789,10 @@
 					else
 						actions.start(new/datum/action/bar/flock_repair(F), user)
 			if(/obj/flock_structure/ghost)
-				actions.start(new /datum/action/bar/flock_deposit(target), user)
+				if (user.resources == 0)
```
stonepillars, line 792:
```suggestion
				if (user.resources <= 0)
```
Just in case resources is negative or something weird.

PR #96, opened 2022-03-26
Add Flock protection component

FlameArrow57:
## About the PR 
This PR adds a backend flock protection component. It's meant to be added to anything Flock owned, and it handles the cases of preventing Flockdrones from attacking the thing it's added to, as well as if attacking the thing is to be reported by a nearby Flockdrone or not.

This is a base component that will likely be modified later. It doesn't currently handle projectile attacks, and it doesn't handle distinguishing between stuff owned by different Flocks.

## Why's this needed? 
For having anything Flock owned have the properties of being immune to Flockdrone attacks and being reported to be attacked automatically by a Flockdrone, this would help make a single place where this sort of code can be modified.

### DIFF
--- a/code/datums/flock/flockprotection.dm
+++ b/code/datums/flock/flockprotection.dm
@@ -43,5 +43,3 @@
 		if (!flockdrone.flock.isEnemy(attacker))
 			flock_speak(flockdrone, "Damage sighted on [source], [pick_string("flockmind.txt", "flockdrone_enemy")] [attacker]", flockdrone.flock)
 			flockdrone.flock.updateEnemy(attacker)
-
-			return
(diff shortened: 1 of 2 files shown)


code/datums/flock/flockprotection.dm
```diff
@@ -0,0 +1,60 @@
...
+			flock_speak(flockdrone, "Damage sighted on [source], [pick_string("flockmind.txt", "flockdrone_enemy")] [attacker]", flockdrone.flock)
+			flockdrone.flock.updateEnemy(attacker)
+
+			return
+
+		if (!attack_message_on_cd)
+			attack_message_on_cd = TRUE
+			SPAWN(attack_message_cd) attack_message_on_cd = FALSE
+
+			flock_speak(flockdrone, "[source] being attacked", flockdrone.flock)
```
stonepillars, line 60: This message probably isn't worth the SPAWN. Machines should report when they take damage, I'd imagine this would get pretty spammy if people started breaking just wall stuff. 

If the flockdrone can see them, and they're an enemy, they would start shooting at them! I'm pretty sure that already makes a message.

PR #97, opened 2022-03-26
Allow Flock structures to report when attacked

FlameArrow57:
[FEATURE]

## About the PR 
This PR just allows flock structures to report when attacked or damaged, including when attacked by non-lethal projectiles such as taser shots.

Ghost structures aren't hit by projectiles at the moment, so projectiles passing through them aren't reported, not sure if that should be or not.

## Why's this needed? 
It allows the Flock to know when its structures are damaged.


code/obj/flock/structure/flock_structure_parent.dm
```diff
@@ -168,13 +171,27 @@
...
+	for (var/mob/living/critter/flock/drone/flockdrone in view(7, src))
+		if (!isdead(flockdrone) && flockdrone.is_npc && flockdrone.flock && flockdrone.flock == src.flock)
+			nearby_flockdrones++
+
+	if (!nearby_flockdrones)
+		if (!attack_message_on_cd)
+			attack_message_on_cd = TRUE
+			SPAWN(30 SECONDS) attack_message_on_cd = FALSE
+
+			flock_speak(src, "ALERT: Under attack", flock)
```
stonepillars, line 193: Flockdrones don't need to see the damage for the flock structure to report it.

```suggestion
	if (!attack_message_on_cd)
		attack_message_on_cd = TRUE
		SPAWN(30 SECONDS) attack_message_on_cd = FALSE

		flock_speak(src, "ALERT: Under attack", flock)
```

Additionally, you should consider moving the flock_speak call onto the flock structure process call. I'd ask pali about whether using a SPAWN for this sort of thing is a good idea.

FlameArrow57: It's taking into consideration Flockdrone reporting of the attack from the Flockdrone protection component, because right now if the reporting feature isn't used and the structure is attacked, the attacker will never be designated as an enemy

stonepillars: Flock structures **should not require a flockdrone to be in view** to report damage. It **should make an alert when damaged**. If a flockdrone sees it, it will also activate the code on the Flockdrone component for witnessing damage, and mark them as hostile accordingly.

Right now **if no flockdrones are nearby** the structure never says anything. It's an expensive loop for no reason.

FlameArrow57: Seems like that'd be unnecessary stuff to check/do every process when it's only needed for when attacked?

stonepillars: Aloe suggests looking into the ON_COOLDOWN macro.

PR #98, opened 2022-03-28
Fill missing pixels in fibrenet sprite

FlameArrow57:
## About the PR 
This PR just fills in the missing pixels in the fibrenet sprite with pixels that have an alpha value of 1.

## Why's this needed? 
It would be annoying to click on a fibrenet sprite and not have the click register, considering the sprite used.

Seems catwalks don't currently have this, not sure if that matters.


stonepillars:
I can't see the icondiff for this yet so scream at me if it breaks anything

PR #100, opened 2022-03-29
Fix being able to fling Flock floor tiles

FlameArrow57:
[BUG]

## About the PR 
This PR fixes being able to fling Flock floor tiles with a crowbar.

## Why's this needed? 
It shouldn't be able to happen with Flock floor tiles needing to be damaged and removed manually with a crowbar.

### DIFF
--- a/code/modules/items/specials/ItemSpecials.dm
+++ b/code/modules/items/specials/ItemSpecials.dm
@@ -1772,14 +1772,17 @@ ABSTRACT_TYPE(/datum/item_special/spark)
 			if (!hit)
 				if (istype(turf,/turf/simulated/floor))
 					var/turf/simulated/floor/F = turf
-					var/obj/item/tile = F.pry_tile(master, user, params)
-					if (tile)
-						hit = 1
-						user.visible_message("<span class='alert'><b>[user] flings a tile from [turf] into the air!</b></span>")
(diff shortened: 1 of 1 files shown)


code/modules/items/specials/ItemSpecials.dm
```diff
@@ -1772,14 +1772,17 @@ ABSTRACT_TYPE(/datum/item_special/spark)
...
-						user.visible_message("<span class='alert'><b>[user] flings a tile from [turf] into the air!</b></span>")
-						logTheThing("combat", user, null, "fling throws a floor tile ([F]) from [turf].")
+					if (istype(F, /turf/simulated/floor/feather))
+						boutput(user, "<span class='alert'><b>The tile stays stuck to the floor!</b></span>")
+					else
+						var/obj/item/tile = F.pry_tile(master, user, params)
+						if (tile)
+							hit = 1
+							user.visible_message("<span class='alert'><b>[user] flings a tile from [turf] into the air!</b></span>")
+							logTheThing("combat", user, null, "fling throws a floor tile ([F]) from [turf].")
```
stonepillars, line 1782: Early return instead of if/else.

stonepillars (changes requested):

stonepillars (approved):

PR #102, opened 2022-03-30
Allow for Flockwalls to be breakable

FlameArrow57:
[FEATURE]

## About the PR 
This PR allows Flockwalls to be breakable

For the basics of breaking a Flockwall, you can hit it with an object enough until it breaks, whereupon you need to use a crowbar to finally destroy the wall, which will create a Flock floortile in its place.

All numbers used were placeholder numbers as they would probably be best decided later in testing.

Some breakable features, such as what happens when you use an RCD, or projectile attacks, were left out as that again would probably be best decided in another PR or in the future.

## Why's this needed? 
It was a planned feature that was not implemented yet.


code/turf/turf_flock.dm
```diff
@@ -250,22 +250,146 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
...
+
+	src.takeDamage("mixed", damage, FALSE)
+	src.visible_message("<span class='alert'>[initial(src.name)] is hit by the blob!/span>")
+
+	if (src.health <= 0)
+		src.destroy()
+
+/turf/simulated/wall/auto/feather/proc/takeDamage(damageType, amount, playAttackSound = TRUE)
+	/*
+	switch(damageType)
```
amylizzle, line 352: Don't leave commented out code in - let git handle the version control

code/turf/turf_flock.dm
```diff
@@ -250,22 +250,146 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
 	desc = "You can feel it thrumming and pulsing."
 	icon = 'icons/misc/featherzone.dmi'
 	icon_state = "0"
-	mat_appearances_to_ignore = list("steel","gnesis")
+	health = 10
+	var/max_health = 10
```
amylizzle, line 254: this seems pretty low - it probably ought to be in line with reinforced wall

amylizzle (changes requested):

amylizzle (approved):

code/turf/turf_flock.dm
```diff
@@ -270,6 +270,7 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
 	..()
 	setMaterial(getMaterial("gnesis"))
 	src.health = src.max_health
+	src.AddComponent(/datum/component/flock_protection, FALSE, TRUE, TRUE)
```
amylizzle, line 273: Are you sure we want to be notified when someone attacks a converted wall? That seems like it would get irritating real quick.

FlameArrow57: It will only be done for new attackers.

PR #107, opened 2022-03-31
Fix cage health

FlameArrow57:
[BUG]

## About the PR 
This PR just fixes a bug with Flock cage health where it could have up to two times its starting value.

## Why's this needed? 
Bug fix. Don't think it would be good to have randomization for Flock cage health.


amylizzle:
seems like this should just be part of #106 and use the maxhealth var you set there

amylizzle:
seems like this should just be part of #106 and use the maxhealth var you set there

PR #108, opened 2022-04-01
Fix Flock stuff having no attack delay

### LINKED ISSUE #78: Flock structures can be destroyed very fast when hitting them
### Labels

[BUG]

### Description

Flock structures can be destroyed fast when hitting them with an object.

### Steps to reproduce

1. Be a human
2. Spam click structure with a fire extinguisher

### Expected Behavior

For there to be a delay between attacks on the Flock structure.

_No response_

FlameArrow57:
[BUG]

## About the PR 
This PR fixes being able to attack Flock structures, Flock floortiles, and Flock cages with every click.

## Why's this needed? 
It wouldn't be good to be able to spam click/attack these.

Fixes #78.


code/obj/flock/structure/flock_structure_parent.dm
```diff
@@ -175,6 +177,8 @@
 
 	takeDamage(damtype, W.force)
 
+	user.lastattacked = src
```
stonepillars, line 180: Why the whitespace?

FlameArrow57: No reason in particular, I'll remove it if that's better

PR #110, opened 2022-04-02
Gnesis turret

amylizzle:
## About the PR 
Adds a turret that shoots syringes full of goagulated gnesis at enemies of the flock

TODO: better sprites

## Why's this needed? 
More building options, and will ideally showcase the basics of event unlocked structures when that's all merged.


code/obj/flock/structure/gnesisturret.dm
```diff
@@ -0,0 +1,147 @@
...
+	var/range = 8
+	var/spread = 1
+	var/datum/projectile/syringe/gnesis/current_projectile = null
+
+	var/powered = FALSE
+	// flockdrones can pass through this
+	passthrough = TRUE
+	usesgroups = TRUE
+	var/fluid_gen_cost = 30 //generating gnesis consumes compute
+	var/base_compute = 20
```
TobleroneSwordfish, line 26: If this is going to take compute then it needs to be really obvious that that's why it's not working, maybe add a message to the examine like "offline due to insufficient compute"

TobleroneSwordfish (changes requested):
Sorry, I did most of this review when you first posted this and forgot to hit submit

stonepillars (approved):
lgtm

PR #112, opened 2022-04-02
Fixes a couple issues with items being awkward

### LINKED ISSUE #1: Drone Pass-the-Parcel with the secure briefcase
### Labels

[BUG]

### Description

Drones will fail to consume secure briefcase, drop it, and the next drone will pick it up

### Steps to reproduce

1. Pop the egg on the bridge
2. Drones find the secure briefcase, try to rummage
3. fail to rummage, drop it
4. another drone picks it up, does the same
5. pass the parcel

### Expected Behavior

maybe containers that fail to open should be blacklisted? or just destroyed?

_No response_

amylizzle:
## About the PR 
Fixes #1 

## Why's this needed? 
It seems the author really wanted to handle picking up items like a player does. I fixed that a bit, and made it so locked or awkward containers get eaten even if they can't open them normally.

### DIFF
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -498,8 +498,9 @@ butcher
 
 /datum/aiTask/succeedable/rummage/succeeded()
 	var/obj/item/storage/container_target = holder.target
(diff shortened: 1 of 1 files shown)


stonepillars (approved):
cursed

PR #114, opened 2022-04-02
Prevent blocking from shocking Flockdrones

FlameArrow57:
[BUG]

## About the PR 
This PR fixes a bug where blocking as a Flockdrone could trigger the anti-grapple system on yourself.

## Why's this needed? 
Bug fix.


amylizzle:
How do you trigger this bug?

FlameArrow57:
You can just possess a Flockdrone, block, and run around a bunch.

amylizzle (approved):

PR #127, opened 2022-04-06
Change how doors are converted

### LINKED ISSUE #17: Converting pod bay doors deletes them
### Labels

[BUG]

### Description

Spraying a turf with pod bay doors on them deletes the pod bay.

### Steps to reproduce

1. Spray nanites on pod bay door
2. See bug

### Expected Behavior

It should replace them with a wall or door-like flockturf

image

FlameArrow57:
[BUG]

## About the PR 
This PR just changes how doors are converted.

It allows podbay doors and door type shutters to be converted to Flockdoors.
These:
image

These (only where connected to space and used as a podbay hangar door or mass driver door):
image

It also removes healing Flockdoors on conversion, which isn't needed since the health will already be full on initial conversion, and the door will be recreated on any future tile conversions.

It also fixes a bug where Flockdoors would be deleted if the tile they were on was not Flock and converted, it fixes windoors being made into Flockdoors, and it fixes #17.

## Why's this needed? 
The changes allow for some bug fixes.


code/datums/flock/flock.dm
```diff
@@ -541,33 +540,35 @@
...
-					converted.set_dir(dir)
-					animate_flock_convert_complete(converted)
+				if (istype(O, /obj/machinery/door))
+					if (istype(O, /obj/machinery/door/firedoor/pyro))
+						qdel(O)
+						continue
+					if (istype(O, /obj/machinery/door/window) || istype(O, /obj/machinery/door/airlock/pyro/glass/windoor) || istype(O, /obj/machinery/door/unpowered/wood))
+						qdel(O)
+						continue
+					if (istype(O, /obj/machinery/door/poddoor/pyro) && !(findtext(O.name, "cargo") || findtext(O.name, "pod") || findtext(O.name, "mass") || findtext(O.name, "disposal"))) // shutters
```
amylizzle, line 552: This cannot be the best way of doing this. Is there a particular reason we can't just convert them all?

code/datums/flock/flock.dm
```diff
@@ -541,33 +540,35 @@
...
-					// if the object has contents, move them over!!
-					for (var/obj/OO in O)
-						OO.set_loc(converted)
-					for (var/mob/M in O)
-						M.set_loc(converted)
-					qdel(O)
-					converted.set_dir(dir)
-					animate_flock_convert_complete(converted)
+				if (istype(O, /obj/machinery/door))
+					if (istype(O, /obj/machinery/door/firedoor/pyro))
```
amylizzle, line 546: Would probably benefit from a comment just explaining that this multi-line if is just for readability. Or you could convert it to a proper multi-line if but those are not so easy to read.

amylizzle, line 546: Would probably benefit from a comment just explaining that this multi-line if is just for readability. Or you could convert it to a proper multi-line if but those are not so easy to read.

amylizzle (changes requested):

code/datums/flock/flock.dm
FlameArrow57: Windoors are found over tables, and for the last case, some of these doors are found in glass windows, such as for the ai core shield, or over tables, such as for the bridge lockdown shutters, meaning if they were converted there would be tables inside doors.

Didn't seem like sliding windoors and sliding windows were big enough for Flock door creation too in terms of making sense.

code/datums/flock/flock.dm
```diff
@@ -541,33 +540,35 @@
...
+						qdel(O)
+						continue
+					if (istype(O, /obj/machinery/door/window) || istype(O, /obj/machinery/door/airlock/pyro/glass/windoor) || istype(O, /obj/machinery/door/unpowered/wood))
+						qdel(O)
+						continue
+					if (istype(O, /obj/machinery/door/poddoor/pyro) && !(findtext(O.name, "cargo") || findtext(O.name, "pod") || findtext(O.name, "mass") || findtext(O.name, "disposal"))) // shutters
+						qdel(O)
+						continue
```
stonepillars, line 554: Maybe just allow `replacement_path` to be `null` and have that signify that that type is just deleted when its found?

code/datums/flock/flock.dm
```diff
@@ -541,33 +540,35 @@
...
+				var/dir = O.dir
+				var/obj/replacementPath = flock_conversion_paths[keyPath]
```
stonepillars, line 556: PATHS ARE NOT OBJECTS
```suggestion
				var/replacementPath = flock_conversion_paths[keyPath]
```

stonepillars (changes requested):

code/datums/flock/flock.dm
FlameArrow57: The issue with using null is that if a Flockdoor is over a tile that isn't a Flock floortile, and you convert that tile, the Flockdoor is deleted, meaning in your conversion list you have to say /obj/machinery/door/feather = /obj/machinery/door/feather, which seemed a bit off, and everything checked for here would also have to be put in the list with a null replacement, which I'm not sure if it's a good idea adding more checks considering its checking everything in the list for every object on a turf. Idk.

code/datums/flock/flock.dm
FlameArrow57: Would work, but seemed like some shutters should be converted, whereas others shouldn't, like chemistry shutters felt like they shouldn't be converted, where ones directly connected to space should. Idk.

PR #134, opened 2022-04-08
Makes flockdrone density work properly

### LINKED ISSUE #132: Flockdrone corpses are dense
### Labels

[BUG]

### Description

Flockdrones are dense even when dead

### Steps to reproduce

1. Flockdrone
2. Kill flockdrone
3. Walk into corpse
4. oh no it's dense and I pushed it instead of standing on it

Not really sure why, but something is constantly resetting density to 1. My first thought is the floor running code, but it didn't seem to be getting called.

### Expected Behavior

Density should be 0 after death.

_No response_

amylizzle:
## About the PR 
Flockdrone corpses no longer block movement or bullets. Also death of a drone ends all lifeprocesses because it ought to, and it was making things dense when they weren't supposed to be.

## Why's this needed? 
Fixes #132

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -217,7 +217,7 @@
 
 /mob/living/critter/flock/drone/Cross(atom/movable/mover)
 	if(isflock(mover))
-		return 1
+		return TRUE
 	else
 		return !src.density
 
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -219,7 +219,7 @@
 	if(isflock(mover))
 		return 1
 	else
-		return 0
+		return !src.density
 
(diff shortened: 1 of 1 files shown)


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -219,7 +219,7 @@
 	if(isflock(mover))
 		return 1
```
stonepillars, line 220:
```suggestion
		return TRUE
```
just while we're here

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -548,6 +560,7 @@
 	..()
 	src.icon_state = "drone-dead"
 	playsound(src, "sound/impact_sounds/Glass_Shatter_3.ogg", 50, 1)
+	src.reduce_lifeprocess_on_death()
 	src.set_density(0)
```
stonepillars, line 564:
```suggestion
	src.set_density(FALSE)
```
here too

PR #135, opened 2022-04-08
Flocktrace abilities (ping)

TobleroneSwordfish:
[FEATURE]
## About the PR 
Removes the place tealprint ability from flocktraces and replaces it with designate enemy and a new ping ability also added to flockminds.

The ping can target any atom, effects are as follows:
- plays a "ping" sound to all sentient members of the flock
- outlines the target with a varying glowy outline
- posts a clickable link to teleport to the target in chat, teleporting to the ping highlights it again
- creates a clientside arrow over every sentient flock member, pointing in the direction of the ping

icon
tree
image
image

The sound effect is a modified version of https://freesound.org/people/Alexhanj/sounds/528730/, licensed under the Creative Commons 0 License.

Known issues:
- Flocktraces can't teleport to pings, despite the `Topic` callback being on the parent class, not sure why.
- The arrow looks kinda silly due to being an animated point sprite.
- Spam pinging.

## Why's this needed? 
Flocktraces need more abilities and ping systems are a tried and tested team communication method.


stonepillars:
Related to #13

code/datums/flock/flock.dm
```diff
@@ -202,19 +201,96 @@
...
+		dummy.appearance_flags = PIXEL_SCALE | RESET_TRANSFORM | RESET_COLOR | PASS_MOUSE
+		dummy.icon = target.icon
+		dummy.icon_state = target.icon_state
+		target.render_target = ref(parent)
+		dummy.render_source = target.render_target
+		dummy.add_filter("outline", 1, outline_filter(size=1,color="#00ff9d"))
+		target.vis_contents += dummy
+
+		SPAWN(0)
+			while(world.timeofday < src.end_time)
```
ZeWaka, line 278: TIME
currently will lead to lots of !FUN! at midnight.

ZeWaka (changes requested):

code/datums/flock/flock.dm
TobleroneSwordfish: !FUN! removed

code/datums/flock/flock.dm
```diff
@@ -202,19 +201,96 @@
...
+
+//is this a weird use case for components? probably, but it's kinda neat
+/datum/component/flock_ping
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+
+	var/const/duration = 5 SECOND
+	var/end_time = -1
+	var/obj/dummy = null
+
+	Initialize()
```
stonepillars, line 256: `RegisterWithParent` should only be called when its actually added to something. I'd ask @pali6 for more.

code/datums/flock/flock.dm
```diff
@@ -202,19 +201,93 @@
...
+/datum/component/flock_ping
+	dupe_mode = COMPONENT_DUPE_UNIQUE
+
+	var/const/duration = 5 SECOND
+	var/end_time = -1
+	var/obj/dummy = null
+
+	RegisterWithParent()
+		//this cast is horribly unsafe, but we just need vis_contents
+		var/atom/movable/target = parent
```
stonepillars, line 258: Override `Initialize` with a check for `ismovable()`, like:
```dm
/datum/component/transfer_output/Initialize()
	if (!isatom(parent))
		return COMPONENT_INCOMPATIBLE
```

code/datums/flock/flock.dm
```diff
@@ -202,19 +201,97 @@
...
+		dummy.add_filter("outline", 1, outline_filter(size=1,color="#00ff9d"))
+		target.vis_contents += dummy
+
+		SPAWN(0)
+			while(TIME < src.end_time)
+				animate(dummy, time = duration/9, alpha = 100)
+				for (var/i in 1 to 4)
+					animate(time = duration/9, alpha = 255)
+					animate(time = duration/9, alpha = 100)
+				sleep(duration)
```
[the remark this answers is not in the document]
TobleroneSwordfish: Ah but we don't know how long to wait for initially since the tile might be re-pinged, so we sleep for the first duration, then check if the time has been reset by another component being added and if it has we extend the animation and sleep again etc.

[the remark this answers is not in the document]
pali6: Hmm the re-ping could cancel the previous animation and start a new one the same way.

amylizzle:
Minor bug: if you ping a floor tile below an object, it makes the object fade in-and-out of visibility repeatedly until the ping wears off. This is probably not actually fixable, but I thought I'd mention it.

code/datums/flock/flock.dm
```diff
@@ -275,24 +275,29 @@
...
+
 		SPAWN(0)
 			while(TIME < src.end_time)
-				animate(dummy, time = duration/9, alpha = 100)
-				for (var/i in 1 to 4)
-					animate(time = duration/9, alpha = 255)
-					animate(time = duration/9, alpha = 100)
-				sleep(duration)
+				var/delta = src.end_time - TIME
+				sleep(min(src.duration, delta))
```
ZeWaka, line 283: why not just spawn for this in the first place lmao - you changed the whole semantics of this, was that intentional?

TobleroneSwordfish: Because the end_time gets modified when it's re-pinged, so it has to sleep for the initial duration and then check if it should sleep for longer. I'm sorry, I don't see how you could do this with just a spawn?
image

TobleroneSwordfish:
> Minor bug: if you ping a floor tile below an object, it makes the object fade in-and-out of visibility repeatedly until the ping wears off. This is probably not actually fixable, but I thought I'd mention it.

I'd lean towards saying this is intentional, because it lets you highlight an object under a lot of other things and have it always render on top. Maybe not great for floor tiles, but since it's always a little transparent I think it's fine.

PR #139, opened 2022-04-10
Tile designation fix

### LINKED ISSUE #27: Highlighting a tile as priority prevents you from clicking on anything on that tile
### Labels

[BUG]

### Description

Making a tile priority prevents you from clicking on stuff on it (namely drones)

### Steps to reproduce

1. Become flockmind
2. Set tile as priority
3. Be unable to click anything on the tile

### Expected Behavior

To be able to click on things (namely drones) on that turf.

Clicking on the turf activates the intangible examine. The turf appears to be _on top_ of everything. Should probably. not do that.

amylizzle:
## About the PR 
Fixes #27 by adding the image to the lighting layer, with a fallback (some tiles don't have lighting) to the old behaviour.


FlameArrow57:
Correct me if I'm wrong, but I think in doing it this way it makes it so that objects can cover up the outline and color of the designation, and it maybe could be a bit hard to see if the lights are turned off, which I think the current way of doing it prevents.

amylizzle:
No you're right, but from messing with it I don't think you can do it as a client image without this tradeoff. The client side image layer seems to determine the click collision, and I don't see an easy way around that.

An alternative that I haven't explored yet is using TobleroneSwordfish's ping highlighting stuff instead. I'm not sure how that'll work here.

amylizzle:
DNM, gonna look at doing some rendering bullshit instead

FlameArrow57:
Is the layer parameter needed in the image() procs? From DM ref, seemed like plane overrides layer?

Also, so this just fixes the issue for tiles that have lighting? What tiles don't have lighting?

code/datums/flock/flock.dm
```diff
@@ -250,11 +250,17 @@
...
-			// create a new image
-			I = image('icons/misc/featherzone.dmi', T, "frontier")
-			I.blend_mode = BLEND_ADD
+			// create a new image and attach it to the light overlay
+			var/iloc
+			if(T.RL_MulOverlay) //if there isn't a lighting layer, fall back to just overlaying the tile
+				iloc = T.RL_MulOverlay
+			else
+				iloc = T
+			I = image('icons/misc/featherzone.dmi', iloc, "frontier", T.layer)
```
FlameArrow57, line 259:
```suggestion
			I = image('icons/misc/featherzone.dmi', T.RL_MulOverlay ? T.RL_MulOverlay : T, "frontier", T.layer)
```

code/datums/flock/flock.dm
```diff
@@ -264,22 +270,28 @@
...
-			// create a new image
-			I = image('icons/misc/featherzone.dmi', T, "frontier")
-			I.blend_mode = BLEND_ADD
+			// create a new image and attach it to the light overlay
+			var/iloc
+			if(T.RL_MulOverlay)
+				iloc = T.RL_MulOverlay
+			else
+				iloc = T
+			I = image('icons/misc/featherzone.dmi', iloc, "frontier", T.layer)
```
FlameArrow57, line 279:
```suggestion
			I = image('icons/misc/featherzone.dmi', T.RL_MulOverlay ? T.RL_MulOverlay : T, "frontier", T.layer)
```

FlameArrow57: If not too hard to read

amylizzle:
> Is the layer parameter needed in the image() procs? From DM ref, seemed like plane overrides layer?
Whoops, that wasn't supposed to be left in - that was from earlier messing around. Will remove.

> Also, so this just fixes the issue for tiles that have lighting? What tiles don't have lighting?
Space tiles don't have lighting, those are the only ones I found, but probably there are others.

PR #142, opened 2022-04-11
Disallow broken walls from reflecting energy projectiles

FlameArrow57:
[FEATURE]

## About the PR 
This PR just makes it so that broken Flockwalls don't reflect energy projectiles.

## Why's this needed? 
Adds another feature to broken Flockwalls besides preventing Flockwalls from floorrunning.

Think it would make a bit more sense too in that a Flockwall riddled with cracks wouldn't reflect an energy projectile very well.

### DIFF
--- a/code/turf/turf_flock.dm
+++ b/code/turf/turf_flock.dm
@@ -359,6 +359,7 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
 		src.desc = "It's broken. You could probably use a crowbar to break the pieces apart."
 		src.broken = TRUE
 		src.UpdateIcon()
+		src.material.setProperty("reflective", 25)
 		if (playAttackSound)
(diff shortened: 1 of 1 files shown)


code/turf/turf_flock.dm
```diff
@@ -359,6 +359,7 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
 		src.desc = "It's broken. You could probably use a crowbar to break the pieces apart."
 		src.broken = TRUE
 		src.UpdateIcon()
+		src.material.setProperty("reflective", 25)
```
stonepillars, line 362: Reading material code this probably changes the material's reflective value. everywhere.

FlameArrow57: I tested reflecting projectiles off other walls after breaking a wall, it doesn't seem to mess with other walls

PR #143, opened 2022-04-11
Change Flockmind death condition

FlameArrow57:
[BUG]

## About the PR 
This PR changes the Flockmind death condition to check for drones and certain structures rather than compute.

Fixes a bug where the Flockmind would still be alive even if all drones were dead.

Also, it structures the Flockmind death check in a way that removes the need for the Flock rift and eggs to have compute and allows future added objects or flock structures to be checked in the death check.

## Why's this needed? 
Bug fix (If the Flockmind loses all its drones but still has sources of compute, it cannot do anything.).

Don't think it makes sense for rift and eggs to provide compute.

### DIFF
--- a/code/mob/living/intangible/flock/flockmind.dm
+++ b/code/mob/living/intangible/flock/flockmind.dm
@@ -58,9 +58,14 @@
 
 /mob/living/intangible/flock/flockmind/Life(datum/controller/process/mobs/parent)
 	if (..(parent))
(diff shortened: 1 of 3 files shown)


FlameArrow57:
Not a feature

amylizzle (approved):
This is a much less hacky way of doing it, :thumbsup:

PR #144, opened 2022-04-11
Allow Flock protection component to handle projectiles

FlameArrow57:
[FEATURE]

## About the PR 
This PR allows the Flock protection component to handle projectile attacks.

Currently projectiles from grenades, like stinger grenades, are not accounted for. Not sure if it is possible to add at the moment.

Shooting a wall with an energy projectile still marks you as an enemy.

## Why's this needed? 
Shooting Flock stuff doesn't currently mark you as an enemy, meaning you can shoot Flock stuff without anything bad happening to you.

### DIFF
--- a/code/datums/flock/flockprotection.dm
+++ b/code/datums/flock/flockprotection.dm
@@ -2,15 +2,18 @@
 	var/flockdrones_can_hit
 	var/report_hand_attack
 	var/report_obj_attack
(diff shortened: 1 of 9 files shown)


amylizzle:
Does this also fix #176 ?

FlameArrow57:
It does not, this only handles the reporting aspect.

PR #148, opened 2022-04-12
Mark people who throw objects at Flockdrones as enemies

FlameArrow57:
[FEATURE]

## About the PR 
This PR makes it so that anyone who throws an object at a Flockdrone, that hits the Flockdrone, is marked as an enemy.

## Why's this needed? 
Currently, hitting a Flockdrone with a thrown object does damage to it, but you aren't marked as an enemy.

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -460,7 +460,7 @@
 /mob/living/critter/flock/drone/hitby(atom/movable/AM, datum/thrown_thing/thr)
 	. = ..()
 	var/mob/attacker = thr.user
-	if(istype(attacker) && !istype(attacker, /mob/living/critter/flock/drone))
+	if(istype(attacker) && !isflock(attacker))
 		src.harmedBy(attacker)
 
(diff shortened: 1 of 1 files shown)


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -457,6 +457,12 @@
 		if(istype(attacker))
 			src.harmedBy(attacker)
 
+/mob/living/critter/flock/drone/hitby(atom/movable/AM, datum/thrown_thing/thr)
+	. = ..()
+	var/mob/attacker = thr.user
+	if(istype(attacker) && !istype(attacker, /mob/living/critter/flock/drone))
```
amylizzle, line 463:
```suggestion
	if(istype(attacker) && !isflock(attacker))
```

FlameArrow57: I only did it that way because I don't think Flockbits can pick up items, but idk

stonepillars: Use isflock

ZeWaka (approved):

stonepillars (approved):

PR #149, opened 2022-04-12
Allow Flock protection component to account for thrown objects

FlameArrow57:
[FEATURE]

## About the PR 
This PR allows the Flock protection component to account for thrown object attacks.

## Why's this needed? 
Throwing objects at Flock stuff does damage, but doesn't mark you as an enemy, meaning you can throw objects at Flock stuff and not be attacked.


code/datums/flock/flockprotection.dm
```diff
@@ -62,6 +65,13 @@
 		else
 			src.attempt_report_attack(source, user)
 
+/datum/component/flock_protection/proc/handle_hitby_thrown(source, atom/hit_atom, datum/thrown_thing/thr)
+	var/mob/attacker = thr.user
+	if(!istype(attacker))
+		return
+	if (!istype(attacker, /mob/living/critter/flock/drone) && report_thrown_attack)
```
stonepillars, line 72:
```suggestion
	if (!isflock(attacker) && report_thrown_attack)
```

PR #150, opened 2022-04-12
Prevent setting signals for Flockdoors with multitools

FlameArrow57:
[BUG]

## About the PR 
This PR fixes a bug where you could set signals for Flockdoors with a multitool.

## Why's this needed? 
Bug fix.

### DIFF
--- a/code/obj/machinery/door/feather.dm
+++ b/code/obj/machinery/door/feather.dm
@@ -13,6 +13,8 @@
 /obj/machinery/door/feather/New()
 	..()
 	setMaterial("gnesis")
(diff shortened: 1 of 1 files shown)


code/obj/machinery/door/feather.dm
```diff
@@ -13,6 +13,8 @@
 /obj/machinery/door/feather/New()
 	..()
 	setMaterial("gnesis")
+	var/datum/component/C = src.GetComponent(/datum/component/mechanics_holder)
+	C.RemoveComponent()
```
stonepillars, line 17:
```suggestion
	C?.RemoveComponent()
```
technically nullable in case this component ever disappears from doors in the future

PR #153, opened 2022-04-13
Absorb resources from items based on amount in stack

### LINKED ISSUE #80: Breaking down a stack of items as Flockdrone provides the same number of resources as one of the items
### Labels

[BUG]

### Description

Breaking down a stack of items only gives the resources of one of the items.

### Steps to reproduce

1. Be Flockdrone
2. Reclaim coil of 30 wires

### Expected Behavior

Resources gained should include resources for all items in the stack.

_No response_

FlameArrow57:
[BUG]

## About the PR 
This PR makes it so that resources gained from reclaimed items are based on the amount of items in the stack.

The speed at which items that are a stack are reclaimed wasn't changed, I wasn't sure how that should be handled.

## Why's this needed? 
Bug fix, fixes #80.


amylizzle:
yeah, I don't think this is the right way to do it. I tried picking up a stack of floor tiles, and it ate them nearly instantly and gave me almost 1000 resources at once.

Maybe instead have the consuming do it's thing and decrement the stack count by one instead.

amylizzle (approved):

amylizzle:
er..github is having a moment, I did not approve this, I approved a different PR :|

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -319,7 +319,7 @@
 	if(I)
 		var/absorb = clamp(src.absorb_rate, 0, I.health)
 		I.health -= absorb
-		src.resources += src.absorb_per_health * absorb
+		src.resources += src.absorb_per_health * absorb * I.amount
```
amylizzle, line 322:
```suggestion
		src.resources += src.absorb_per_health * absorb 
	        I.amount--
```

amylizzle (changes requested):
change to eat one item at a time

code/mob/living/critter/flock/flockdrone.dm
amylizzle: note - this might be part of the fix, but it is not the whole fix

amylizzle:
> Yeah, I was able to get it so that the total time it takes to eat a stack of items is the sum of the times of each individual item, but might be kind of hard to remove individual items from the stack, I would have to spend more time on that. The whole stack has one health value.

It has an amount though, you could grab the health value, tick it down, when it's <=0 decrease item count, and reset health to max again?

stonepillars:
Playing cards are `W_CLASS_TINY`, and so are their boxes. It may be better to rebalance based on `I.w_class`
If say, TINY items only gave 1 resource, then drones could only really get 52 resources from a deck of cards.
Stuff like cable coils are both tiny, and have an amount value, but unlike card decks do not have contents so drones can't
drop them in that way.

**If you want to just let the flock eat card decks all at once**, then just check if there's an amount and consume based on that, and then also ignore the contents check since you just ate it all at once.

If they're balanced based on that, then drones can easily prioritize yummier items (bigger items) and then gather around the pile of cards on the floor like a flock of birds like they're supposed to when they're done.

stonepillars:
Closed for design reasons. Very difficult to reconcile the fact when an item has a near-zero health it still retains full functionality which can give the flock an advantage in some circumstances. The serial approach will be used instead.

PR #163, opened 2022-04-15
Flock protection comp blah

stonepillars:
## About the PR 

Slaps #149 and #144 together and does a bunch of cool refactoring to the protection component.

## Why's this needed? 

Normalizes both changes because it would be confusing to merge both
without named parameters and other fun stuff that would've
caused problems.

### github magic words

close #149 
close #144


code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,92 @@
...
+		return
+	if (!isflock(attacker) && src.report_proj)
+		src.attempt_report_attack(source, attacker)
+
+/// Look for flockdrones nearby who will snitch
 /datum/component/flock_protection/proc/attempt_report_attack(source, mob/attacker)
-	var/list/nearby_flockdrones = list()
+	var/mob/living/critter/flock/drone/snitch
 	for (var/mob/living/critter/flock/drone/flockdrone in view(7, source))
 		if (!isdead(flockdrone) && flockdrone.is_npc && flockdrone.flock)
```
amylizzle, line 83:
```suggestion
		if (!isdead(flockdrone) && flockdrone.is_npc && flockdrone.flock && flockdrone.flock == source.flock)
```

amylizzle (changes requested):

code/datums/flock/flockprotection.dm
stonepillars: Would not compile, source may not be flock. Probably should require a ref to the flock datum in the init

stonepillars: Turns out all of the flock furniture does not keep a reference to the flock datum. Which means either I make it have one or we just assume flocks protect other flocks.

code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,92 @@
...
-				boutput(user, "<span class='alert'>The grip tool refuses to allow damage to this, jamming briefly.</span>")
-				return TRUE
-		else
+	var/we_take_those = FALSE
+	if (istype(source, /obj/lattice/flock) && isweldingtool(W) && W:try_weld(user,0,-1,0,0))
+		we_take_those = TRUE
+	else if (istype(source, /obj/storage/closet/flock))
+		var/obj/storage/closet/flock/flockcloset = source
+		if (!flockcloset.open)
+			if(!isflock(user) && istype(W, /obj/item/cargotele))
```
FlameArrow57, line 47:
```suggestion
			if(!istype(W, /obj/item/cargotele))
```
Could probably give the message to flock too since it isn't harm

code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,92 @@
...
 
+/// Protect against punches/kicks/etc.
 /datum/component/flock_protection/proc/handle_attackhand(source, mob/user as mob)
 	if (user.a_intent != INTENT_HARM)
 		return
 
-	if (istype(user, /mob/living/critter/flock/drone) && !flockdrones_can_hit)
-		if (istype(source, /obj/storage/closet/flock) || istype(source, /turf/simulated/floor/feather) || istype(source, /obj/machinery/door/feather))
-			return
+	if (isflock(user) && src.flock_immune)
```
FlameArrow57, line 32: Means message will be given even though no damage can be done for the closet, turf, and door, when humans can use harm intent with no damage or report given. Not sure how much it matters.

code/datums/flock/flockprotection.dm
FlameArrow57: Not sure if flock furniture needs it as it doesn't really have health? Think people deconstructing flock furniture could be viewed as a repurposing/replacement type of thing?

stonepillars: Decoupled protection from flock being interested in protecting it. Enables multiple flocks and variant configuration of the hit detection

code/obj/flock/furniture.dm
```diff
@@ -235,7 +235,7 @@
 /obj/lattice/flock/New()
 	..()
 	setMaterial("gnesis")
-	src.AddComponent(/datum/component/flock_protection, FALSE, FALSE, TRUE)
+	src.AddComponent(/datum/component/flock_protection)
```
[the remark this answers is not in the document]
stonepillars: But punching them is still sort of mean. Resolving because of refactor (recreate in correct place)

code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,92 @@
...
-				boutput(user, "<span class='alert'>The grip tool refuses to allow damage to this, jamming briefly.</span>")
-				return TRUE
-		else if (istype(source, /obj/machinery/door/feather))
-			var/mob/living/critter/flock/drone/flockdrone = user
-			if (flockdrone.find_type_in_hand(/obj/item))
-				boutput(user, "<span class='alert'>The grip tool refuses to allow damage to this, jamming briefly.</span>")
-				return TRUE
-		else
+	var/we_take_those = FALSE
+	if (istype(source, /obj/lattice/flock) && isweldingtool(W) && W:try_weld(user,0,-1,0,0))
```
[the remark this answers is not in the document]
stonepillars: Should be fixed I think?

stonepillars:
Possibly blocked by atom property for flock converted things. Would avoid a bunch of crap with type checks during claiming

amylizzle (approved):

code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,81 @@
...
+		return COMPONENT_INCOMPATIBLE
+
+	src.flock = flock
+
+/datum/component/flock_interest/RegisterWithParent()
+	RegisterSignal(parent, COMSIG_FLOCK_ATTACK, .proc/handle_flock_attack)
+
+/// If flockdrone is in our flock, deny the attack, otherwise scream and cry
+/datum/component/flock_interest/proc/handle_flock_attack(source, mob/living/critter/flock/user)
+	if(!ismob(user))
```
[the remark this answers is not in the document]
stonepillars: Breaks the entire component. I made the type that to avoid a re-cast later on but can change to a cast later on

code/datums/flock/flockprotection.dm
```diff
@@ -1,76 +1,81 @@
...
+		return FALSE
+
+	if (istype(user) && user.flock == src.flock)
+		boutput(user, "<span class='alert'>The grip tool refuses to harm this, jamming briefly.</span>")
+		return TRUE
+
+	var/mob/living/critter/flock/drone/snitch
+	for (var/mob/living/critter/flock/drone/FD in view(7, source))
+		if (!isdead(FD) && FD.is_npc && FD.flock == src.flock)
```
amylizzle, line 27:
```suggestion
		if (!isdead(FD) && FD.is_npc && FD.flock == src.flock && FD != user)
```

code/datums/flock/flockprotection.dm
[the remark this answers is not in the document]
amylizzle: oh, I think it's probably better for readability to cast it later on.

code/datums/flock/flockprotection.dm
amylizzle: wait no that's dumb

PR #181, opened 2022-04-16
Flockdrone menu

TobleroneSwordfish:
[FEATURE] [QOL]
## About the PR 
Adds context actions Flockminds and Flocktraces can use for flockdrone manual control. A Flockmind or Flocktrace can click on a drone to order it to:

- Move somewhere
- Convert something
- Capture somebody
- Build a barricade
- Shoot somebody

image
Clockwise from top: convert, move, shoot, barricade, capture.
Upon selecting any of the options, your cursor turns into an ability cast and you can then click anywhere to specify the target.

Let me know if there's any other tasks that should be here, or tasks that are here but shouldn't be, and also if you think Flocktraces should have access to this or not.
## Why's this needed? 
Finer control over drones is extremely useful for micromanaging your flock.
+1 step towards true high APM RTS flock gameplay.


amylizzle (commented):
This is really good stuff. Gonna playtest it a bit to see if I can break it, but on the face of things it looks great :+1:

amylizzle:
Could use a message on fail/success of receiving order. "That's not a valid target" or something

amylizzle:
Opening the menu and then moving doesn't update the position of the menu on your screen - should probably follow the drone if at all possible

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -905,3 +921,110 @@ butcher
...
+			holder.interrupt()
+
+	on_reset()
+		..()
+		holder.target = get_turf(src.target)
+
+/datum/aiTask/timed/targeted/flockdrone_shoot/targetable
+	switched_to()
+		on_reset()
+		if (!ismob(src.target))
```
amylizzle, line 1022:
```suggestion
		if (!ismob(src.target) && !iscritter(src.target))
```

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -835,6 +850,7 @@ butcher
 			else if(!actions.hasAction(owncritter, "flock_entomb")) // let's not keep interrupting our own action
 				if(owncritter.active_hand != 2) // nanite spray
 					owncritter.set_hand(2)
+				if (owncritter.a_intent != INTENT_DISARM)
```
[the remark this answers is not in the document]
TobleroneSwordfish: No actually, it was originally all under a check for active_hand which was causing some bugs, but it should really just be in switched_to or on_reset anyways. Will fix.

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -64,6 +68,9 @@
 			emote("beep")
 			say(pick_string("flockmind.txt", "flockdrone_created"))
 
+	for (var/type in childrentypesof(/datum/contextAction/flockdrone))
```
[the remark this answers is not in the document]
TobleroneSwordfish: Well it's not a `/datum/contextAction/flockdrone`, it's a type, but I'll add `as anything`

TobleroneSwordfish:
> Opening the menu and then moving doesn't update the position of the menu on your screen - should probably follow the drone if at all possible

Sadly as far as I can see that would require completely re-implementing contextActions, since they're hud elements rn.

amylizzle:
> > Opening the menu and then moving doesn't update the position of the menu on your screen - should probably follow the drone if at all possible
> 
> Sadly as far as I can see that would require completely re-implementing contextActions, since they're hud elements rn.

Yeah, I figured it'd be impossible.

amylizzle:
Bug: you can order a drone to shoot another drone and it'll work. The drone you targeted is designated as enemy and you cannot undesignate it because it's not a valid target.

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -905,3 +922,118 @@ butcher
...
+			holder.interrupt()
+
+	on_reset()
+		..()
+		holder.target = get_turf(src.target)
+
+/datum/aiTask/timed/targeted/flockdrone_shoot/targetable
+	switched_to()
+		on_reset()
+		if (!ismob(src.target) && !iscritter(src.target))
```
amylizzle, line 1029:
```suggestion
		if (!(ismob(src.target) || iscritter(src.target) || isflock(src.target)))
```

TobleroneSwordfish: I think you mean `if (!(ismob(src.target) || iscritter(src.target)) || isflock(src.target))`, so I'm going to commit that

amylizzle (approved):

PR #183, opened 2022-04-17
Make every health-based Flock thing repairable

FlameArrow57:
[FEATURE]

## About the PR 
This PR adds repair functionality to everything Flock that doesn't have repair functionality, except for furniture and lights as they aren't health based.

The health repair numbers used are placeholders.

## Why's this needed? 
You should be able to repair damaged Flock stuff.


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -768,15 +768,15 @@
 			return
 
 	// CONVERT TURF
-	if(!isturf(target) && !(istype(target, /obj/storage/closet/flock) || istype(target, /obj/table/flock) || istype(target, /obj/structure/girder) || istype(target, /obj/machinery/door/feather) || istype(target, /obj/flock_structure/ghost)))
+	if(!isturf(target) && !(istype(target, /obj/flock_structure) || istype(target, /obj/machinery/door/feather) || istype(target, /obj/window/feather) || istype(target, /obj/grille/flock) || istype(target, /obj/storage/closet/flock) || istype(target, /obj/table/flock) || istype(target, /obj/structure/girder)))
```
TobleroneSwordfish, line 771: `HAS_ATOM_PROPERTY(target, PROP_ATOM_FLOCK_THING)`
blame amy if this is wrong

code/mob/living/critter/flockcritter_parent.dm
```diff
@@ -354,46 +354,59 @@
...
+			var/mob/living/critter/flock/flockcritter = target
+			flockcritter.HealDamage("All", flockcritter.health_brute / 3, flockcritter.health_burn / 3)
+			if (flockcritter.is_npc)
+				flockcritter.ai.interrupt()
+		else if (istype(target, /obj/flock_structure))
+			var/obj/flock_structure/structure = target
+			structure.repair()
+		else
+			switch (target.type)
+				if (/obj/machinery/door/feather)
```
TobleroneSwordfish, line 390: This is kinda cursed, but the only way to fix it would be to componentize repair or use the crime `:` operator so it's probably fine

TobleroneSwordfish (changes requested):

code/mob/living/critter/flockcritter_parent.dm
stonepillars: A whole lot in flock can be componentized :shelterfrog;

PR #187, opened 2022-04-17
Adds a tick() call to interrupt

TobleroneSwordfish:
## About the PR 
This just makes it so that interrupt immediately starts the next task instead of waiting for the next AI tick cycle. 

## Why's this needed? 
Makes flockdrones react significantly faster in combat (and when manually ordered by #181), allowing a decent number of them an actual chance at caging someone trying to punch them.
I'm scared that there's a reason that it didn't already work like this, but I can't see one for now.

### DIFF
--- a/code/mob/living/critter/ai.dm
+++ b/code/mob/living/critter/ai.dm
@@ -98,6 +98,7 @@ var/list/ai_move_scheduled = list()
 			current_task?.reset()
 			current_task = default_task
 			stop_move()
+			tick()
 
 	proc/die()
 		src.enabled = 0


amylizzle (approved):
What's the worst that can happen?

TobleroneSwordfish:
> What's the worst that can happen?

exaaactly, it didn't throw a recursion depth runtime in my 10 minutes of testing so I trust it with my life

stonepillars:
If this shit creates a server crash I will throw you under the adminbus

PR #190, opened 2022-04-18
Fix Flock conversion bugs

### LINKED ISSUE #17: Converting pod bay doors deletes them

FlameArrow57:
[BUG]

## About the PR 
This PR allows podbay doors to be converted to Flock doors and fixes various other conversion bugs.

It allows podbay doors and some door type shutters to be converted to Flockdoors.
These:
image

These (only where connected to space and used as a podbay hangar door or mass driver door):
image

It also removes healing Flockdoors on conversion, which isn't needed since the health will already be full on initial conversion, and the door will be recreated on any future tile conversions.

It also fixes Flockdoors being deleted if the tile they were on was not Flock and converted, it fixes windoors being made into Flockdoors, it fixes #17, and it fixes converting various instances of machinery to compute nodes that shouldn't be converted.

Partial reopening of #127, with some more stuff.

## Why's this needed? 
Some changes allow for bug fixes, other stuff is direct bug fixes.

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -656,7 +656,7 @@
 						qdel(O)
 						break
 				if (istype(O, /obj/machinery/computer))
-					if (istype(O, /obj/machinery/computer/card/portable) || istype(O, /obj/machinery/computer/security/wooden_tv) || istype(O, /obj/machinery/computer/secure_data/detective_computer) || istype(O, /obj/machinery/computer/airbr) || istype(O, /obj/machinery/computer/tour_console) || istype(O, /obj/machinery/computer/arcade))
+					if (istype(O, /obj/machinery/computer/card/portable) || istype(O, /obj/machinery/computer/security/wooden_tv) || istype(O, /obj/machinery/computer/secure_data/detective_computer) || istype(O, /obj/machinery/computer/airbr) || istype(O, /obj/machinery/computer/tour_console) || istype(O, /obj/machinery/computer/arcade) || istype(O, /obj/machinery/computer/tetris))
 						break
 				if (istype(O, /obj/machinery/light/lamp) || istype(O, /obj/machinery/computer3/generic/personal) || istype(O, /obj/machinery/computer3/luggable))
 					break
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -577,8 +577,7 @@
 /var/list/flock_conversion_paths = list(
 	/obj/grille/steel = /obj/grille/flock,
 	/obj/window = /obj/window/feather,
-	/obj/machinery/door/airlock = /obj/machinery/door/feather,
-	/obj/machinery/door = null,
+	/obj/machinery/door = /obj/machinery/door/feather,
 	/obj/stool = /obj/stool/chair/comfy/flock,
 	/obj/table = /obj/table/flock/auto,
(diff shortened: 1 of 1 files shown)


code/datums/flock/flock.dm
```diff
@@ -644,36 +643,41 @@
...
-							var/obj/storage/S = O
-							if(!isnull(S.spawn_contents))
-								S.make_my_stuff()
-						// if the object has contents, move them over!!
-						for (var/obj/OO in O)
-							OO.set_loc(converted)
-						for (var/mob/M in O)
-							M.set_loc(converted)
+						break
+					if (istype(O, /obj/machinery/door/poddoor/pyro) && !(findtext(O.name, "cargo") || findtext(O.name, "pod") || findtext(O.name, "mass") || findtext(O.name, "disposal"))) // shutters
```
amylizzle, line 655: I still really don't like relying on `findtext(name, "blah")` - it assumes that mappers have stuck to an unofficial naming scheme, and haven't made a mistake. 

I think a better way would be to check whether the turf is dense without the door, and if it is delete it - otherwise convert it.

code/datums/flock/flock.dm
```diff
@@ -644,36 +643,41 @@
...
 						qdel(O)
-						converted.set_dir(dir)
-						animate_flock_convert_complete(converted)
-					break //we found and converted the type, don't convert it again
-
+						break
+				if (istype(O, /obj/machinery/computer))
+					if (istype(O, /obj/machinery/computer/card/portable) || istype(O, /obj/machinery/computer/security/wooden_tv) || istype(O, /obj/machinery/computer/secure_data/detective_computer) || istype(O, /obj/machinery/computer/airbr) || istype(O, /obj/machinery/computer/tour_console) || istype(O, /obj/machinery/computer/arcade) || istype(O, /obj/machinery/computer/tetris))
```
amylizzle, line 659: If we're not converting these, you should add them to the `flock_conversion_paths` with a null, or if you don't want to delete them, maybe have `/obj/path = FALSE` mean "don't convert or delete this".

Also I get wooden_tv, but why not convert the others?

code/datums/flock/flock.dm
```diff
@@ -644,36 +643,41 @@
...
+						break
+				if (istype(O, /obj/machinery/light/lamp) || istype(O, /obj/machinery/computer3/generic/personal) || istype(O, /obj/machinery/computer3/luggable))
+					break
+			var/dir = O.dir
+			var/replacementPath = flock_conversion_paths[keyPath]
+			var/obj/converted = new replacementPath(T)
```
amylizzle, line 665: for the don't covert/delete this, you could just do `if isnull(replacementPath) delete it else if(!replacementPath) break`

amylizzle (approved):

amylizzle (changes requested):
I pressed the wrong button lol

code/datums/flock/flock.dm
ZeWaka: i'd rather just fix this outright on the upstream by making it a new type or subtype or whatever

amylizzle: that seems like a much nicer fix

PR #192, opened 2022-04-18
Adds flock cheatmode admin verb

TobleroneSwordfish:
[QOL]
## About the PR 
Adds a debug admin verb to toggle cheatmode on a flock, also makes cheatmode give you a million total compute.

## Why's this needed? 
Makes debug easier, can always revert it later to avoid spamming the debug pane.

### DIFF
--- a/code/modules/admin/admin_verbs.dm
+++ b/code/modules/admin/admin_verbs.dm
@@ -2218,14 +2218,14 @@ var/list/fun_images = list()
 
 /client/proc/flock_cheat()
 	SET_ADMIN_CAT(ADMIN_CAT_DEBUG)
-	set name = "Flock cheatmode"
-	set desc = "Toggle cheatmode on or off on a particular flock"
+	set name = "Flock infinite compute"
+	set desc = "Toggle infinite compute on or off on a particular flock"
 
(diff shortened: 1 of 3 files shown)


amylizzle:
maybe have a different cheat for free compute?

TobleroneSwordfish:
> maybe have a different cheat for free compute?

sure

code/modules/admin/admin_verbs.dm
```diff
@@ -2215,3 +2215,17 @@ var/list/fun_images = list()
 				if((i++ % 5) == 0)
 					sleep(1 SECOND)
 				apc.setStatus("lightsout", dur SECONDS)
+
+/client/proc/flock_cheat()
+	SET_ADMIN_CAT(ADMIN_CAT_DEBUG)
+	set name = "Flock infinite compute"
+	set desc = "Toggle infinite compute on or off on a particular flock"
+
+	var/flockname = tgui_input_list(src, "Pick a flock", "Choose flock", flocks)
```
amylizzle, line 2224: Could this also have another menu for which cheat to enable/disable - if we're doing multiple?

Mostly I'm just thinking we should have a cheat for "unlock all the buildings" as well as a cheat for "infinite compute".

amylizzle (approved):

PR #193, opened 2022-04-18
Fixes flockmind being unable to act while partitioning

### LINKED ISSUE #168: Flockmind is unable to do anything after partitioning.
### Labels

[BUG]

### Description

When partitioning you are unable to use any abilities or open the menu until the 30 seconds are up.

### Steps to reproduce

1. Be flock
2. Partition mind
3. Press any ability

### Expected Behavior

I expected to be able to keep doing stuff.

_No response_

TobleroneSwordfish:
[QOL]
## About the PR 
Fixes #168

## Why's this needed? 
Partitioning shouldn't give flockminds a 30 second APM break

### DIFF
--- a/code/mob/living/intangible/flock/flockmind.dm
+++ b/code/mob/living/intangible/flock/flockmind.dm
@@ -118,7 +118,7 @@
 
 
 /mob/living/intangible/flock/flockmind/proc/partition()
-	boutput(src, "<span class='notice'>Partitioning initiated. Stand by.</span>")
+	boutput(src, "<span class='flocksay'>Partitioning initiated. Stand by.</span>")
 
 	var/ghost_confirmation_delay = 30 SECONDS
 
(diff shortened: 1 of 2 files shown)


code/mob/living/intangible/flock/flockmind.dm
```diff
@@ -138,7 +138,7 @@
 	if (!length(candidates))
 		message_admins("No ghosts responded to a Flocktrace offer from [src.real_name]")
 		logTheThing("admin", null, null, "No ghosts responded to Flocktrace offer from [src.real_name]")
-		boutput(src, "<span class='alert'>Unable to partition, please try again later.</span>")
+		flock_speak(null, "Partition failure: unable to coalesce sentience.", src.flock)
```
FlameArrow57, line 141: Not sure if the whole flock needs to know that a partition failed, when it's announced to the flock if a partition is successful, in make_flocktrace(). If kept, should probably change the initial message to tell the flock when a partition is started.

TobleroneSwordfish: good point, I'll change it back to local

PR #197, opened 2022-04-20
Change how annotations are updated

FlameArrow57:
## About the PR 
This PR changes how annotations are done. Now, instead of using a single proc to update every annotation on changing one annotation, annotations are only updated when wanted.

Also makes it so that you can't schedule a busy tile as a priority tile or convert an already reserved turf to help with this.

Fixes a related bug where a turf could be repeatedly reserved.

## Why's this needed? 
It prevents checking annotation updates for things that you aren't looking to update.

It also makes it easier to put multiple annotations on an atom.

It makes it easier to add an annotation that is based on a parameter.


code/datums/flock/flock.dm
```diff
@@ -489,22 +447,35 @@
...
+	I.plane = PLANE_ABOVE_LIGHTING
+	src.annotations_busy_tiles[T] = I
+	src.addClientImage(I)
 
 /datum/flock/proc/unreserveTurf(var/name)
+	var/turf/simulated/T = src.busy_tiles[name]
 	src.busy_tiles -= name
-	src.updateAnnotations()
+
+	var/image/I = src.annotations_busy_tiles[T]
```
amylizzle, line 465: Should probably make sure `istype(T)` here, just in case some bad call to `unreserveTurf(bad name)` or something fucky happens with turf deletion.

amylizzle (changes requested):

code/datums/flock/flock.dm
FlameArrow57: I think ideally this would be handled by onUpdate() of /datum/action/bar/flock_convert?

TobleroneSwordfish:
Something I just realised is I don't think this handles new flocktraces being added to the flock, which is a thing that happens pretty often. They need to be able to see who is marked as enemy etc.

PR #199, opened 2022-04-23
Prevent multiple of the same name from being given to Flock entities

FlameArrow57:
[BUG][FEATURE]

## About the PR 
This PR just prevents Flockminds having the same name as each other, and prevents Flocktraces, Flockdrones, and Flockbits from having the same name of another of their type in their Flock.

The chances of more than one entity having the same name are low enough that the way that was used to do this hopefully shouldn't be an issue.

## Why's this needed? 
Currently some of the code uses names of drones and bits for converting turf. Prevents any bugs that could occur if drones or bits were to have the same name, or naming is used in another way later on.

Gives a unique name to each Flockmind and entity in their Flock.


code/datums/flock/flock.dm
```diff
@@ -392,6 +392,40 @@
...
+
+/datum/flock/proc/pick_name(flock_type)
+	var/name
+	switch(flock_type)
+		if ("flock")
+			while (!name || (locate(name) in flocks))
+				name = "[pick(consonants_lower)][pick(vowels_lower)].[pick(consonants_lower)][pick(vowels_lower)]"
+		if ("flocktrace")
+			var/trace_found = FALSE
+			while (!trace_found)
```
amylizzle, line 405: Potentially infinite loop, if the flock is very big (9025 entities). It's unlikely to come up, but it's better to protect against it anyway.

code/datums/flock/flock.dm
```diff
@@ -392,6 +392,40 @@
 			if(C)
 				C.images -= I
 
+// naming
+
+/datum/flock/proc/pick_name(flock_type)
+	var/name
+	switch(flock_type)
+		if ("flock")
```
amylizzle, line 400: Do we really need to do it by type? If anything it'd probably be better if each one was globally unique, rather than just unique-for-drones or unique-for-traces

code/mob/living/critter/flock/flockbit.dm
```diff
@@ -20,7 +20,7 @@
 		src.zone_sel.change_hud_style('icons/mob/flock_ui.dmi')
 
 	src.name = "[pick_string("flockmind.txt", "flockbit_name_adj")] [pick_string("flockmind.txt", "flockbit_name_noun")]"
-	src.real_name = "[pick(consonants_upper)].[rand(10,99)].[rand(10,99)]"
+	src.real_name = src.flock ? src.flock.pick_name("flockbit") : name
```
amylizzle, line 23: probably better to just make `flock_pick_name()` a global function than to have it as part of the flock datum, since it doesn't actually rely on any flock data.

amylizzle (changes requested):

stonepillars:
Maybe for flockdrones and flocktraces, scope the name to the flock?
I.e. a flocktrace belonging to flock `pa.li` would be `pa.li.po`

code/datums/flock/flock.dm
FlameArrow57: Confused, do you mean unique across flocks?

amylizzle: I just mean it's probably better to get rid of the switch, and just do 
```
for(var/mob/living/F in mobs)
if(isflock(F))
if(F.name == name)
found = TRUE
break
```

So a given flock entity is guaranteed to have a unique name in the game, not just a unique name for all the flockdrones in that flock. Hit me up on discord if it's not clear.

FlameArrow57: Yeah, that make sense, maybe sounds a bit more of a design decision that we should talk about though?

amylizzle: After discord discussion it turns out I'm blind. Consider this resolved. but maybe also consider doing the istype() thing instead of strings

ZeWaka (commented):
could just be a hash table indicating usage but i think this is fine, above would be cool though

stonepillars:
image

FlameArrow57:
Would rather new naming scheme be done in another PR, since this was meant just for something else, and that is still in discussion I think?

amylizzle (changes requested):

code/datums/flock/flock.dm
```diff
@@ -351,6 +351,50 @@
...
 
+// naming
+
+/datum/flock/proc/pick_name(flock_type)
+	var/name
+	var/name_found = FALSE
+	var/tries = 0
+	var/max_tries = 5000 // really shouldn't occur
+	switch(flock_type)
+		if ("flock")
```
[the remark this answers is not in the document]
amylizzle: I just noticed you committed this and then reverted it, I'm guessing because you made it an if-chain instead of a switch?

stonepillars (commented):
The if chain has a lot of copy+pasted repetition. A mapping from name to thing might be better for finding if anything shares that name, then you can just have a short if that generates the names and loop until you find something or hit maximum retries.
```dm
if(name in src.units_by_name)
    // ...
 ```
is better than
```
for (var/mob/concrete/type/here/F in src.units)
``` and finding the names yourself, and esp. with concrete types it can still lead to name duplication since you do not check outside of that type of unit.

**Comment by FlameArrow57** [comment]
Idk, that would kind of mean looping over names that you know couldn't be common, none of the flock things share the same naming scheme. Plus, flocks is a global variable, so would still have to loop through for flockmind naming.

File: code/datums/flock/flock.dm [hunk 5]
```diff
@@ -351,6 +351,50 @@
...
+		if ("flockbit")
+			while (!name_found && tries < max_tries)
+				name = "[pick(consonants_upper)].[rand(10,99)].[rand(10,99)]"
+				tries++
+				for (var/mob/living/critter/flock/bit/F in src.units)
+					if (F.name == name)
+						break
+				name_found = TRUE
+	if (!name_found && tries == max_tries)
+		return "error"
```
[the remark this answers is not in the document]
FlameArrow57: I added a debug message. Idk about hardcoding something in, if this breaks something would have had to gone seriously wrong and there would be more broken than this.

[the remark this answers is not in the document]
amylizzle: yeah I guess, just generally not good to show obvious errors to users if you can avoid it. You could always do "er.ro.r" :P

stonepillars:
> Idk, that would kind of mean looping over names that you know couldn't be common, none of the flock things share the same naming scheme. Plus, flocks is a global variable, so would still have to loop through types for flockmind naming.

```dm
for (var/T as anything in list)
	if (T.name == name) break
```
This pattern is repeated alot in the code when it could be a list of all currently used names by flock things.
This means **nothing, not even across types** will have a duplicate name. You won't have drones
with duplicate names to flockminds or traces, etc.
It would also be a hash lookup instead of iterating across everything.

Not sure where `flocks` came up

code/datums/flock/flock.dm
```diff
@@ -358,18 +360,56 @@
...
+	return name
+
 // UNITS
 
 /datum/flock/proc/registerUnit(var/atom/movable/D)
 	if(isflock(D))
 		src.units |= D
+		if (src.active_unit_names[D:real_name])
+			D:real_name = istype(D, /mob/living/critter/flock/drone) ? src.pick_name("flockdrone") : src.pick_name("flockbit")
```
stonepillars, line 404: I'd rather the name be put into the list by `pick_name`.
Either way, get rid of `:`. There is already a type check you can cast and use dot.

stonepillars (changes requested):

code/datums/flock/flock.dm
```diff
@@ -358,18 +360,56 @@
 	for (var/mob/M in src.annotation_viewers)
 		M.client?.images -= I
 
+// naming
+
+/datum/flock/proc/pick_name(flock_type)
+	var/name
+	var/name_found = FALSE
```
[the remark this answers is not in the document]
FlameArrow57: You mean using break instead of name_found = TRUE?

Also, I don't think datums have the name var?

code/datums/flock/flock.dm
FlameArrow57: It already is put in by pick_name?

TobleroneSwordfish: it's indexed by real_name though, which is not guaranteed to be unique or static

code/datums/flock/flock.dm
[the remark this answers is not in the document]
stonepillars: Nevermind, datums don't have names. your right.
As for name_found, I'm saying that if you are using an associative list you don't need it, since the check "is this name used" is a single expression (`src.active_unit_names[name]`)

code/datums/flock/flock.dm
FlameArrow57: Sorry, just to clarify, this just repicks the flock unit's name if its name is already found in the units, otherwise no naming happens. Atm, all flock units are registered before their names are picked, except for one case which I accounted for. Was meant mainly for if wanting to undormantize a flock unit, thing for the future I think, but can be removed if wanted.

pick_name should guarantee it to be unique, not sure if possible to account for static name changes, but I think ideally a drone's real_name shouldn't be changing

code/datums/flock/flock.dm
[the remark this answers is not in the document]
FlameArrow57: I think the while() loop would break on first try if using !src.active_unit_names[name]?

amylizzle (approved):

stonepillars (approved):
good now unless im blind

PR #205, opened 2022-04-24
Add toggleable deconstruction

amylizzle:
## About the PR 
Clicking a thing marked for deconstruction will now stop it being deconstructed and interrupt drones who are on their way to do that.

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -453,6 +453,12 @@
 		count++
 	return count
 
+/datum/flock/proc/toggleDeconstructionFlag(var/atom/target)
+	if(target in src.deconstruct_targets)
+		src.deconstruct_targets -= target
+	else
+		src.deconstruct_targets += target
+	src.updateAnnotations()
 // ENEMIES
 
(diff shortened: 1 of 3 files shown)


code/datums/flock/flock.dm
```diff
@@ -453,6 +453,12 @@
 		count++
 	return count
 
+/datum/flock/proc/toggleDeconstructionFlag(var/atom/target)
+	if(target in src.deconstruct_targets)
+		src.deconstruct_targets -= target
+	else
+		src.deconstruct_targets += target
+	src.updateAnnotations()
```
stonepillars, line 461: I'm certain pali or waka will complain that this isn't a macro

amylizzle: is there a macro for this?

stonepillars: No but proc calls are slow or something :shelterfrog:

FlameArrow57: togglePriorityTurfs() does this, plus see annotations changes I made for togglePriorityTurfs(), would be making similar changes once this is merged

PR #212, opened 2022-04-26
Remove Flock critters that leave the station z level from their Flock

FlameArrow57:
[FEATURE]

## About the PR 
This PR removes Flockdrones and Flockbits that leave the station z level from the Flock, by dormantizing them.

Currently, stopping the Flockbit animation doesn't work, not sure why, and Flockbits need a dormant sprite. The dormant sprite used here is an unanimated version of the regular Flockbit sprite.

Doesn't handle dormancy checks anywhere else than for clicking on links to teleport to Flockdrones in the Flock radio.

## Why's this needed? 
They shouldn't count towards the Flock if leaving the z level, or do anything on other z levels. This is just one implementation of that.


stonepillars:
bird go sleep

PR #213, opened 2022-04-27
Moves the target reset from sequence to aiTask

TobleroneSwordfish:
[BUG]
## About the PR 
Moves the `holder.target = null` so it applies to all aiTasks by default

## Why's this needed? 
Fixes flockdrones trying to deconstruct walls with incapacitors and is probably just a good idea in general

### DIFF
--- a/code/mob/living/critter/ai.dm
+++ b/code/mob/living/critter/ai.dm
@@ -204,7 +204,6 @@ var/list/ai_move_scheduled = list()
 	proc/next_task()
 		return null
 
-	SHOULD_CALL_PARENT(TRUE)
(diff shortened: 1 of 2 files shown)


code/mob/living/critter/ai.dm
```diff
@@ -205,6 +205,7 @@ var/list/ai_move_scheduled = list()
 		return null
 
 	proc/on_reset()
+		holder.target = null
```
amylizzle, line 208:
```suggestion
		SHOULD_CALL_PARENT(TRUE)
		holder.target = null
```

PR #222, opened 2022-04-29
Prevent drones from reporting dormancy on deletion

FlameArrow57:
[BUG]

## About the PR 
This PR fixes a bug where Flockdrones could report the dormant message for going out of range if sent to nullspace (I think, didn't test).

## Why's this needed? 
Bug fix.


amylizzle (approved):

stonepillars:
bird go where no bird hath gone b4

PR #224, opened 2022-04-29
Fix being able to use cargo teleporters on Flock closets

FlameArrow57:
[BUG]

## About the PR 
This PR fixes being able to use the cargo teleporter on Flock closets.

## Why's this needed? 
Bug fix.


amylizzle (approved):
That is a weird way of managing a list of allowed types. This seems like the best solution without rejigging it all.

PR #230, opened 2022-04-30
Standardize material across Flock stuff

FlameArrow57:
[FEATURE][BUG]

## About the PR 
This PR standardizes how material properties are set across all Flock stuff. It gives Flock stuff that didn't have the gnesis material the gnesis material, prevents some Flock stuff that had the gnesis material from showing that in its name and description, and fixes chairs and alcoves giving steel sheets when deconstructed.

Only thing I noticed was that Flock windows are now opaque. Wasn't sure if opacity should be set manually or the material itself changed.

## Why's this needed? 
Consistency and bug fixes.

### DIFF
--- a/code/obj/window.dm
+++ b/code/obj/window.dm
@@ -1059,7 +1059,6 @@
 
 /obj/window/feather/New()
 	..()
-	setMaterial(getMaterial("gnesis"))
 	src.AddComponent(/datum/component/flock_protection, FALSE, TRUE, TRUE)
 
 /obj/window/feather/special_desc(dist, mob/user)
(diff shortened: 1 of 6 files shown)


TobleroneSwordfish:
another thing I've noticed is that the starting rift and eggs are tinted due to being made of gnesis, I think things that already have flock sprites probably shouldn't be permanently tinted by their material

code/obj/window.dm
```diff
@@ -1050,12 +1050,16 @@
...
+	mat_appearances_to_ignore = list("gnesis")
+	mat_changename = FALSE
+	mat_changedesc = FALSE
 	health = 50 // as strong as reinforced glass, but not as strong as plasmaglass
 	health_max = 50
 	density = 1
 
 /obj/window/feather/New()
 	..()
+	setMaterial(getMaterial("gnesis"))
```
amylizzle, line 1062: why not just leave it at the default_material of gnesisglass?

stonepillars, line 1062:
```suggestion
```

PR #243, opened 2022-05-01
Add Flockmind promotion framework

FlameArrow57:
[FEATURE]

## About the PR 
This PR adds procs in that may be used elsewhere for allowing the Flockmind to promote Flocktraces to Flockmind.

## Why's this needed? 
It would be nice for the Flockmind to be able to give another entity control of the flock if they need to leave or go afk.

### DIFF
--- a/code/mob/living/intangible/flock/flocktrace.dm
+++ b/code/mob/living/intangible/flock/flocktrace.dm
@@ -68,6 +68,33 @@
 		stat("Flock:", "none")
 		stat("Drones:", 0)
 
+/mob/living/intangible/flock/trace/proc/promoteToFlockmind(remove_flockmind_from_flock)
+	var/was_in_drone = FALSE
+	var/mob/living/critter/flock/drone/controlled = src.loc
+	if (istype(controlled))
+		was_in_drone = TRUE
+		controlled.release_control(FALSE)
+
+	boutput(src, "<span class='flocksay'><b>\[SYSTEM: New functions detected. Control of Flock assumed.\]</b></span>")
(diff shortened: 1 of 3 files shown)


stonepillars:
Probably should be clearer when using the ability that it you will become a ghost and lose control of your flock

FlameArrow57:
There's the option to just become the trace that you promoted instead, it depends on the parameter.

code/mob/living/intangible/flock/flocktrace.dm
```diff
@@ -68,6 +68,33 @@
...
+	else
+		var/mob/living/intangible/flock/flockmind/F = new (get_turf(src), src.flock)
+		src.mind.transfer_to(F)
+		if (was_in_drone)
+			controlled.take_control(F, FALSE)
+		src.flock.removeTrace(src)
+		src.flock.hideAnnotations(original)
+		original.ghostize()
+		qdel(original)
+		qdel(src)
```
stonepillars, line 96: Could remove the negation and swap the code blocks around like this?
```suggestion
	if (remove_flockmind_from_flock)
		var/mob/living/intangible/flock/flockmind/F = new (get_turf(src), src.flock)
		src.mind.transfer_to(F)
		if (was_in_drone)
			controlled.take_control(F, FALSE)
			src.flock.removeTrace(src)
			src.flock.hideAnnotations(original)
			original.ghostize()
			qdel(original)
			qdel(src)
	else
		src.mind.swap_with(original)
		if (was_in_drone)
			src.set_loc(get_turf(original))
			controlled.take_control(original, FALSE)
```

PR #245, opened 2022-05-01
shut up you stupid birds

amylizzle:
## About the PR 
Makes flock beeps quieter, so you can be sneakier when you deploy.

## Why's this needed? 
Play test revealed that the flock could be heard from half the station away. this is bad.


stonepillars:
silence burds

PR #246, opened 2022-05-01
Flock relay now blocks shuttle calls and recalls the shuttle if it is in transit

TobleroneSwordfish:
[BALANCE] [FEATURE]
## About the PR 
When the relay is constructed it recalls the shuttle if it is on its way to the station and blocks any further shuttle calls. When it is destroyed it allows the shuttle to be called once again, and when it activates it automatically calls an unrecallable shuttle.

## Why's this needed? 
Crew should not be able to call shuttle to escape the relay, they should attempt to destroy it instead.
See https://github.com/stonepillars/goon-flock/discussions/239.

### DIFF
--- a/code/datums/gamemodes/flock.dm
+++ b/code/datums/gamemodes/flock.dm
@@ -4,7 +4,6 @@
 
 	shuttle_available = 2
 	shuttle_available_threshold = 12000 // 20 min, default value, probably change this
-	shuttle_unavailable_msg = "CentCom reports extreme anomalous radio signals originating from your station. The shuttle has been returned to base as a precaution."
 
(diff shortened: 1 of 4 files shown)


code/datums/gamemodes/flock.dm
```diff
@@ -4,6 +4,7 @@
 
 	shuttle_available = 2
 	shuttle_available_threshold = 12000 // 20 min, default value, probably change this
+	shuttle_unavailable_msg = "CentCom reports extreme anomalous radio signals originating from your station. The shuttle has been returned to base as a precaution."
```
stonepillars, line 7: Since shuttle_available is always `2` for this game mode, this message most likely never appears.

PR #249, opened 2022-05-01
Make flocktraces be antags

### LINKED ISSUE #235: Flocktraces don't get the antag mark
### Labels

[BUG]

### Description

Flocktraces aren't reported as antags 

### Steps to reproduce

1. Have a bunch of flocktraces
2. None of them have the antag "V" symbol in admin view
3. round ends
4. not listed as antags

### Expected Behavior

They should be added to the antag list

_No response_

amylizzle:
## About the PR 
Anyone who gets flocktrace'd will now be added to the antag list and recieve the little V symbol for admins

## Why's this needed? 
Fixes #235

### DIFF
--- a/_std/defines/roles.dm
+++ b/_std/defines/roles.dm
@@ -33,6 +33,7 @@
 // gimmicks
 #define ROLE_BATTLER "battler"
 #define ROLE_FLOCKMIND "flockmind"
+#define ROLE_FLOCKTRACE "flocktrace"
 #define ROLE_OMNITRAITOR "omnitraitor"
 #define ROLE_GRINCH "grinch"
 #define ROLE_FLOOR_GOBLIN "floor_goblin"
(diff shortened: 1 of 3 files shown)


FlameArrow57:
Is new the right place for this? I think maybe it would be better to put in the make flocktrace procs and stuff based on how other antagonists are handled but not sure.

amylizzle:
I think so. It's not like there are ever going to be flocktraces that are not antags, and traces can't exist independent of flocks, so `New()` will always be called by a flock datum. If I put it somewhere else, it'll make future code need to copy it to multiple places and that is bad.

stonepillars:
New() *is* a weird place to put it. There is the evilize proc and other ways antags are done. I'd ask in #imcoder about what the current best practice is for it.

amylizzle:
I was just looking at where else I might put it, and the flock datum's `registerTrace` function is literally called only in flocktrace's `New()` 

also there's an unused `spawnTrace()` function for some reason.

amylizzle:
ooh it could maybe go in `/mob/proc/make_flocktrace()`?

PR #253, opened 2022-05-02
Give Flockdrones and Flockbits health icons

FlameArrow57:
[FEATURE]

## About the PR 
This PR gives Flockdrones and Flockbits health icons that show their damage levels.

In game:

healthicons

States:

image

## Why's this needed? 
It'd be nice to see the more precise damage levels of drones other than what's shown by their damaged states.


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -674,6 +681,28 @@
 				src.icon_state = "drone-d2"
 	return
 
+/mob/living/critter/flock/drone/proc/update_health_icon()
```
TobleroneSwordfish, line 684: Should probably move this down to `/mob/living/critter/flock` instead of copy pasting

FlameArrow57: I was uncertain because saying src.update_health_icon() doesn't work in some cases in /mob/living/critter/flock, such as in New() with the new names for Flockdrones and Flockbits going to be requiring their real_name set based on the Flock. Would kind of have some cases where src.update_health_icon() can't be written in some procs for /mob/living/critter/flock and needs to be individually set in the Flockdrone and Flockbit files, but the proc is still written in the /mob/living/critter/flock file.

amylizzle: maybe don't have it in the flockdrone/bit, but instead with the other annotation stuff? That way you could just iterate over `flock.units` instead?

amylizzle:
This looks really good, but I have a couple points:

- it should be toggleable. I probably won't want to see this all the time. I wonder if maybe it could be done when alt is held down like with names?
- This might be a good place to do the indicator for "a flocktrace is controlling this drone". Maybe with a little highlight or extra icon on the plumbob?

TobleroneSwordfish:
> * This might be a good place to do the indicator for "a flocktrace is controlling this drone". Maybe with a little highlight or extra icon on the plumbob?

It's not going to be very noticeable though, especially if health is only displayed while holding alt. I think it'd probably be better as a separate icon, but we accumulating a lot of overhead icons at this point so idk.

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: Okay better question, why are we using `real_name` there anyways and not just ref? Using a (technically) mutable name var as a key is questionable.

TobleroneSwordfish: There's a builtin byond `ref(object)` proc which just returns a guaranteed unique string for any object, I think that's a better way of doing it?

stonepillars:
!merge_upstream

code/mob/living/critter/flock/flockdrone.dm
FlameArrow57: Seems to be working now using src, not sure why

stonepillars:
!merge_upstream

ZeWaka:
> It's not going to be very noticeable though, especially if health is only displayed while holding alt.

but it's not?

TobleroneSwordfish:
> but it's not?

from the comment I was replying to:
> * I wonder if maybe it could be done when alt is held down like with names?

PR #256, opened 2022-05-04
Make Flockwalls only be on when a Flockdrone is floorrunning through them

FlameArrow57:
## About the PR 
This PR just makes it so that Flockwalls are only on when a Flockdrone is floorrunning through them.

Ex. comparison:
wall off

wall on

## Why's this needed? 
It should help reduce some of the visual load that can happen with so many animated walls on the Flockmind's screen.

Also a way of improving floorrunning feedback for non-Flock players as you know when a Flockdrone is floorrunning through a wall.


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -487,14 +475,12 @@
 				var/turf/simulated/wall/auto/feather/wall = src.loc
 				if (wall.broken)
 					return ..()
+				if (!wall.on)
+					wall.on()
```
TobleroneSwordfish, line 479: This looks spooky, but testing with medass shows they don't conflict since DM doesn't treat methods as properties

TobleroneSwordfish (approved):

code/mob/living/critter/flock/flockdrone.dm
stonepillars: based non-first-class-functions

PR #257, opened 2022-05-05
Add Flock plating under Flock floortiles

FlameArrow57:
[FEATURE]

## About the PR 
This PR adds a new Flock tile, a plating, meant to be inactive, not associated with any Flock. It is found under Flock floortiles.

image

It acts like typical plating, where you can put a floor tile on it to make it into a seemingly regular tile. Prying the floor tile off will reveal the Flock plating underneath.

## Why's this needed? 
It makes more sense for cases like when you build a Flock floortile in space, it's broken and pried up, and a Flock plating appears rather than a typical station plating tile.

It also further reinforces the idea that Flock conversion of a tile is full conversion.


stonepillars:
I don't really subscribe to the idea that it is a total conversion esp. on flock tiles- and these would be covered by either flock floors or normal floor tiles.

code/turf/turf.dm
```diff
@@ -742,6 +756,9 @@ proc/generate_space_color()
 		floor.icon_state = icon_old
 	return floor
 
+/turf/proc/ReplaceWithFeatherPlating()
```
ZeWaka, line 759: Why is this its own snowflake proc

code/turf/turf.dm
```diff
@@ -46,6 +46,7 @@
 	var/tmp/allow_unrestricted_hotbox = 0
 	var/wet = 0
 	var/sticky = FALSE
+	var/flock_plating_under = FALSE
```
ZeWaka, line 49: Doc Comments.

code/turf/turf_flock.dm
```diff
@@ -3,6 +3,79 @@
 // -----
 // FLOOR
 // -----
+/turf/simulated/floor/feather_plating
```
ZeWaka, line 6: Why is this not plating

ZeWaka (changes requested):

code/turf/turf.dm
FlameArrow57: Was for convenience, will remove

code/turf/turf.dm
FlameArrow57: Let me know if needs to be improved

code/turf/turf_flock.dm
FlameArrow57: Done

ZeWaka (approved):

stonepillars:
!merge_upstream

PR #258, opened 2022-05-05
Make Flockdrone abilities unresumable

### LINKED ISSUE #247: Cancelling a floor conversion by clicking on another tile cancels animation but not the action
### Labels

[MINOR BUG]

### Description

The tile animation stops and the progress bar continues, action finishes successfully.

### Steps to reproduce

1. Start converting a floor as a drone
2. Click on another floor

### Expected Behavior

Either nothing or conversion cancelling or starting converting the new tile instead.

Likely the reason is the action being resumable or something like that.

FlameArrow57:
[BUG]

## About the PR 
This PR just makes the Flockdrone abilities unresumable, based on what Pali suggested in the linked issue.

I guess this now makes some sound spam possible, which could be accounted for with cooldowns, but not sure if that was the best solution considering the sound isn't much of a problem for when you aren't intentionally doing that.

## Why's this needed? 
Bug fix, fixes #247.

Also fixes some related ability exploits possible by starting one of these abilities on a tile and ending it while on another tile and spam clicking over the distance, allowing you to do stuff such as converting a tile to flock from several tiles away.

### DIFF
--- a/code/mob/living/critter/flockcritter_parent.dm
+++ b/code/mob/living/critter/flockcritter_parent.dm
@@ -167,6 +167,7 @@
 	id = "flock_convert"
 	interrupt_flags = INTERRUPT_MOVE | INTERRUPT_ACT | INTERRUPT_STUNNED | INTERRUPT_ACTION
 	duration = 45
+	resumable = FALSE
 
 	var/turf/simulated/target
 	var/obj/decal/decal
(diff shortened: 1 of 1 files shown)


stonepillars:
Noise spam is definitely possible, so give some thought to whether adding a cooldown is any different from a user experience perspective.

PR #259, opened 2022-05-06
Fixes drones still moving when controlled

### LINKED ISSUE #248: Taking control of a drone sometimes makes it start moving in some direction
### Labels

[BUG]

### Description

Need to tap the direction to stop

### Steps to reproduce

1. Take control of a drone under ✨some circumstances✨ (presumably when it's walking somewhere?)
2. Watch it move until you tap that directional key

### Expected Behavior

Move stops when you take control of it

_No response_

TobleroneSwordfish:
[BUG]
## About the PR 
Fixes #248, also possibly fixes a bug I found while testing this where drones would get stuck with a zero length path after releasing control over them.

## Why's this needed? 
bug bad + if we fix all pali's bug reports maybe they'll come back to us 🙏


FlameArrow57:
Nice, just checking for a few more cases of walk(src, 0), seems there's two more cases in the flockdrone file, then two cases in the flockbit file. Maybe could be replaced with this? Wasn't sure if needed.

TobleroneSwordfish:
> Nice, just checking for a few more cases of walk(src, 0), seems there's two more cases in the flockdrone file, then two cases in the flockbit file. Maybe could be replaced with this? Wasn't sure if needed.

Most of those seem to be death cases so should be fine, I've replaced one call in the flockbit login code with a `stop_move`.

PR #262, opened 2022-05-06
Fixes flockdrones not being able to put items into lockers

### LINKED ISSUE #105: Flockdrones can't put items in storage containers
### Labels

[BUG]

### Description

Flockdrones can't put items in storage containers like crates and lockers.

### Steps to reproduce

1. Be a Flockdrone
2. Pick an item up
3. Left click a storage container

### Expected Behavior

For the item to be put into the container rather than dropped at the Flockdrone's location.

_No response_

TobleroneSwordfish:
[BUG]
## About the PR 
Fixes #105, technically this is an upstream bug but it's so minor we may as well just lump it in with flock? Can move this to upstream if needed.

## Why's this needed? 
bug

### DIFF
--- a/code/mob/living/critter.dm
+++ b/code/mob/living/critter.dm
@@ -1206,7 +1206,7 @@
 		return O
 
 	drop_item()
-		..()
+		. = ..()
 		src.update_inhands()
 
 	proc/on_sleep()


FlameArrow57:
Nice! Yeah, might be good to upstream pr, as it might mean that ghost critters can now put objects in containers?

TobleroneSwordfish:
alright I PRed it upstream: https://github.com/goonstation/goonstation/pull/8480
feel free to close this if/when that gets merged

FlameArrow57:
Cool, there was also some code in the flock closet file that was used to fix this I think for Flock closets, maybe could be removed instead in this PR or done later in the cleanup PR.

TobleroneSwordfish:
fixed upstrem

PR #265, opened 2022-05-06
quick fix to not move bits onto dense tiles

### LINKED ISSUE #242: Flock Bytes Can Spawn In Flock Walls
### Labels

[BUG]

### Description

When you split a flock drone into flock bytes next to a flock wall, there is a chance the flock byte will move into the flock wall

### Steps to reproduce

1. Get a flock drone
2. Move it next to a flock wall
3. Split it
4. Watch a flock byte move into the wall

### Expected Behavior

For flock bytes to not enter flock walls

_No response_

amylizzle:
## About the PR 
Fixes #242

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -763,6 +763,9 @@
 	// get candidate places to move them
 	var/turf/T = get_turf(src)
 	var/list/candidate_turfs = getneighbours(src)
+	for(var/turf/n in candidate_turfs)
+		if(is_blocked_turf(n))
+			candidate_turfs -= n
 	for(var/i=1 to num_bits)
 		B = new(get_turf(src), F = src.flock)
 		src.flock?.registerUnit(B)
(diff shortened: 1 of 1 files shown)


FlameArrow57:
Does blocked turf take into account density? I think one thing that was different when the new function started being used was that flockbits could start clipping into each other whereas they would all go in different directions before.

Bit more on this, maybe an issue if a flockdrone is walled off on all sides and is split?

amylizzle:
> Does blocked turf take into account density? I think one thing that was different when the new function started being used was that flockbits could start clipping into each other whereas they would all go in different directions before.
> 
> Bit more on this, maybe an issue if a flockdrone is walled off on all sides and is split?

Yeah, `is_blocked_turf` queries the density of the turf and all the atoms in the turf. If any of them is dense it returns `TRUE`.

If there are no valid turfs (because surrounded by walls or whatever), it just spawns all the bits on the turf the drone was on. Or it should, I think now it might actually runtime - I should fix that.

PR #266, opened 2022-05-06
Super Fast AI

amylizzle:
## About the PR 
EDIT: Actually maybe we do merge this? It'll need a couple tweaks to things like absorber timing if we do though.
Okay, I'm not saying we merge this, but you definitely have to try it.

## Why's this needed? 
image

### DIFF
--- a/_std/defines/mob.dm
+++ b/_std/defines/mob.dm
@@ -17,6 +17,7 @@
 #define USR_DIALOG_UPDATES_RANGE 8192	//updateusrdialog will consider this mob as being able to 'attack_ai' and update its ui at range
 #define UNUSED_16384 16384
 #define SHOULD_HAVE_A_TAIL 32768 //Would we miss our tail if it comes off?
+#define HEAVYWEIGHT_AI_MOB 65536 //ai gets ticked every 0.16 seconds instead of the usual 1.6 seconds - gotta go fast
 
 //mob intent type defines
 #define INTENT_HARM "harm"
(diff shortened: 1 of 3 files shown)


stonepillars:
awesome

_std/defines/mob.dm
```diff
@@ -17,6 +17,7 @@
 #define USR_DIALOG_UPDATES_RANGE 8192	//updateusrdialog will consider this mob as being able to 'attack_ai' and update its ui at range
 #define UNUSED_16384 16384
 #define SHOULD_HAVE_A_TAIL 32768 //Would we miss our tail if it comes off?
+#define HEAVYWEIGHT_AI_MOB 65536 //ai gets ticked every 0.16 seconds instead of the usual 1.6 seconds - gotta go fast
```
ZeWaka, line 20: X<<Y please 🥺🥺🥺

code/datums/controllers/process/mob_ai.dm
```diff
@@ -3,7 +3,7 @@
 datum/controller/process/mob_ai
 	setup()
 		name = "Mob AI"
-		schedule_interval = 1.6 SECONDS
+		schedule_interval = 0.16 SECONDS
```
ZeWaka, line 6: Why not a number divisible by 60?

_std/defines/mob.dm
amylizzle: I did all those flags, because I was there.

amylizzle:
Moved upstream goonstation/goonstation#8563

PR #267, opened 2022-05-07
Move shared Flock critter stuff to parent procs

FlameArrow57:
## About the PR 
This PR just moves some stuff shared between Flockdrones and Flockcritters to parent procs.

## Why's this needed? 
Cleanliness.

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -84,7 +84,6 @@
 		if (controller)
 			src.release_control_abrupt()
 		flock_speak(null, "Connection to drone [src.real_name] lost.", src.flock)
-		src.flock.removeDrone(src)
 	src.remove_simple_light("drone_light")
 	..()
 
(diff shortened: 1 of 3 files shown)


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -199,22 +198,20 @@
...
 	src.remove_simple_light("drone_light")
 
 	if (!src.flock)
+		..()
 		return
 
 	src.flock.hideAnnotations(src)
-	src.flock.removeDrone(src)
 
 	if (src.controller)
```
amylizzle, line 211: why not replace this whole chunk with 
```
if(src.controller)
		src.release_control()
```

amylizzle (changes requested):

code/mob/living/critter/flock/flockdrone.dm
FlameArrow57: You'll be free to roam in another z level if not. Would have to click on a name link in chat to get back.

amylizzle: `release_control()` should probably have a z-level check added into it then

FlameArrow57: life() already has the z level check, and calls this

amylizzle: I'm confused. If the problem is that `release_control()` doesn't check for z level, but this function is only called after a check to z level, then the problem is gone?

But also, `release_control()`, if it doesn't already, should detect if the trace/mind is going to be spawned in a non-station z-level and return it to a late join location if it is. Then this whole if block can be reduced to the suggested change.

FlameArrow57: Sorry, to clarify more, this was more implemented for unique stuff, like saying connection to drone lost and messaging the flock that a drone has been lost.

life() checks for z level. If drone is off z level, then dormantize() is called, where it checks to see if the flockmind is still in the drone. If it is, then it will move them back to the station automatically. I realize now that I guess you could eject drone right before that happens and be off z level, but another pr can be made for that.

Flockmind can't go off the station z level as far as I'm aware other than going through a drone, and the rift is anchored, so don't think it's possible for them to be spawned off z level.

amylizzle: Okay, I think I see. Call this resolved then :+1:

PR #268, opened 2022-05-07
Flock walls in 3/4

amylizzle:
## About the PR 
Adds 8dir 3/4ths perspective flock walls. Also makes the windows 3/4ths perspective too, using the material colour of gnesis glass over existing window sprites.

Also, they aren't animated anymore because that was a lot of effort. I might do it later, but maybe we shouldn't because now the frame lag is gone.

Flockvision:
image

Human vision:
image

## Why's this needed? 
Old walls were old and flat.


code/turf/turf_flock.dm
```diff
@@ -269,25 +269,29 @@ turf/simulated/floor/feather/proc/bfs(turf/start)//breadth first search, made by
...
+	flags = USEDELAY | ALWAYS_SOLID_FLUID | IS_PERSPECTIVE_FLUID
 	mat_appearances_to_ignore = list("steel", "gnesis")
 	mat_changename = FALSE
 	mat_changedesc = FALSE
-	connects_to = list(/turf/simulated/wall/auto/feather, /obj/machinery/door/feather)
-
+	connect_overlay = TRUE
+	connect_diagonal = TRUE
+	connects_to = list(/turf/simulated/wall/auto/feather, /obj/machinery/door, /obj/window)
+	connects_with_overlay = list(/obj/machinery/door, /obj/window)
```
FlameArrow57, line 284: Do we want it connecting to non-Flock stuff? Felt it gave more of a contrast when it didn't and makes a bit more sense that it only connects to its own stuff, but if it looks nicer?

code/obj/window.dm
```diff
@@ -1044,6 +1044,28 @@
 
 
 // Flockdrone BS goes here - cirr
+/obj/window/auto/feather
+	default_material = "gnesisglass"
```
FlameArrow57, line 1048: Was wondering if auto windows somehow inherit from the regular windows, but it seems some stuff is missing like ignoring material appearances?

code/obj/window.dm
```diff
@@ -1044,6 +1044,28 @@
...
+
+/obj/window/auto/feather/special_desc(dist, mob/user)
+  if(isflock(user))
+    return {"<span class='flocksay'><span class='bold'>###=-</span> Ident confirmed, data packet received.
+    <br><span class='bold'>ID:</span> Fibrewoven Window
+    <br><span class='bold'>System Integrity:</span> [round((src.health/src.health_max)*100)]%
+    <br><span class='bold'>###=-</span></span>"}
+    // todo: damageable walls
+  else
+    return null // give the standard description
```
FlameArrow57, line 1063: Some indentation stuff that can be fixed here, and comments removed probably

FlameArrow57:
Any reason why Flock trader glass windows aren't changed to auto? Feel like it would be easier to only need to maintain one type of window

code/turf/turf_flock.dm
amylizzle: Yeah, I considered having them separate but it looked shit so I changed it. The connectors round it off nicely.

code/obj/window.dm
amylizzle: willfix

FlameArrow57:
Although if meant to be temporary, I really think some visual effect should still be added for broken and on walls, otherwise not going to be able to tell either of these. Could maybe add a version of each wall with the broken overlay in the featherzone file over each if it doesn't look too bad, and for on walls could just maybe make the runes brighter without flashing?

FlameArrow57:
Also had some minor sprite thoughts. Can be done later though, up to you

FlameArrow57:
> > Any reason why Flock trader glass windows aren't changed to auto? Feel like it would be easier to only need to maintain one type of window
> 
> Short version: I did that initially and it looked terrible.

Ok, in that case, I propose that maybe we move all the variables and stuff from /obj/window/flock to /obj/window/auto/flock, reduce stuff from /obj/window/flock, and make it so that Flock only repairs and deconstructs the auto window, considering they will never see /obj/window/flock?

code/obj/window.dm
amylizzle: window/auto inherits from /obj/window yes, but it overrides a lot of behaviour

code/obj/window.dm
FlameArrow57: Err I mean previous layout was fine, except maybe some lines after return could be indented further, but none of the lines have proper tabbing

amylizzle:
> Although if meant to be temporary, I really think some visual effect should still be added for broken and on walls, otherwise not going to be able to tell either of these. Could maybe add a version of each wall with the broken overlay in the featherzone file over each if it doesn't look too bad, and for on walls could just maybe make the runes brighter without flashing?

I agree, but I think we should do it in another PR because I don't want to do it now and it's good for things to be atomized.




> Ok, in that case, I propose that maybe we move all the variables and stuff from /obj/window/flock to /obj/window/auto/flock, reduce stuff from /obj/window/flock, and make it so that Flock only repairs and deconstructs the auto window, considering they will never see /obj/window/flock?

I guess? It doesn't really matter all that much IMO. I reckon we leave it and just remove it in the future if it becomes a hassle to maintain, which would be surprising given how little behaviour it has.

code/obj/window.dm
amylizzle: oh gross, I didn't even notice

PR #208, opened 2022-04-25
Flock serial stack eating

amylizzle:
## About the PR 
Eat stacks an item at a time, using item health to determine how long it takes to eat one item. After the item is eaten, health of the stack is reset to the initial value.

Competes with #153 - do not merge both. 
Fixes #80

PR #270, opened 2022-05-08
Rework and improve item absorption

### LINKED ISSUE #80: Breaking down a stack of items as Flockdrone provides the same number of resources as one of the items

FlameArrow57:
## About the PR 
This PR reworks and improves item absorption.

Makes variable names more intuitive.

Resources gained per item health absorbed are reduced.

Item eating is now serial based.

Items are stopped from burning on equip.

Rounding is done for absorbed resources.

Resources are no longer given from absorption completion.

Messages for absorbing items are modified to reduce chat spam.

Alternative pr to #208.

Fixes #80.

## Why's this needed? 
More clear variable names are nice.

Resources gained reduced based on playtesting.

Item stacks shouldn't be dissolved like a single item.

Wouldn't want an item being burned at the same time as being reclaimed.

Rounding checks prevent fractional resources.

Doesn't make sense to have resources given on item absorption, given that the item is supposed to be absorbed already.

Chat spam reduction.

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -407,17 +407,17 @@
 		return
 
 	var/absorb = min(src.absorb_rate, I.health)
-	if (absorber.instant_absorb)
+	if (absorber.instant_absorb && !absorber.ignore_amount)
 		boutput(src, "<span class='alert'>[I] is weak enough that it breaks apart instantly!</span>")
 		src.resources += round(src.absorb_per_health * absorb * I.amount)
 	else
 		I.health -= absorb
 		src.resources += round(src.absorb_per_health * absorb)
-		if (I.health > 0 || (I.health == 0 && I.amount > 1))
+		if (I.health > 0 || (I.health == 0 && I.amount > 1 && !absorber.ignore_amount))
 			playsound(src, "sound/effects/sparks[rand(1, 6)].ogg", 50, 1)
 		if (I.health > 0)
 			return
-		if (I.amount > 1)
+		if (I.amount > 1 && !absorber.ignore_amount)
 			if (initial(I.health))
 				I.health = initial(I.health)
 			else
@@ -1092,6 +1092,7 @@
 	icon = 'icons/mob/flock_ui.dmi'
 	icon_state = "absorber"
 	var/instant_absorb = FALSE
+	var/ignore_amount = FALSE
 
 /datum/equipmentHolder/flockAbsorption/can_equip(var/obj/item/I)
 	if (istype(I, /obj/item/grab))
@@ -1104,6 +1105,7 @@
 
 	var/mob/living/critter/flock/drone/F = holder
 	src.instant_absorb = item.amount > 1 && round(F.absorb_per_health * item.health) == 0
+	src.ignore_amount = istype(item, /obj/item/spacecash)
 
 	item.inventory_counter?.show_count()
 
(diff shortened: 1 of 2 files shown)


amylizzle (approved):
Just that tiny balance change, but otherwise looks good.

amylizzle:
There is one *minor* bug, which is that item health gets basically permanently reduced by being digested.
For example

1. Pick up a new item stack (it has 100 health, 80 items)
2. Digest most of the first item (it now has 5 health, 80 items)
3. take it out of the absorbtion holder
4. put it back in
5. now the stack's max health is 5, and you lose 95% of the resources from the stack.

I'm not suggesting it's necessary to fix this (unless you can think of a good way to do so), but just pointing it out.

FlameArrow57:
Yeah will fix, remembered about it a bit ago, think it can be fixed with just putting stuff done in New() for items into a proc

amylizzle:
Note: we need money stacks to get eaten while ignoring amount, because it's possible to have stacks of like a million, and that is a lot of flockdrones.

FlameArrow57:
Ok well the issue was fixed, but it will mean that a stack of items burned to 0.1 health, if picked up, will then have its health reset if picked up by a Flockdrone. Not sure if there's really a way to fix this.

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -27,8 +27,7 @@
 	butcherable = 1
 
 	var/absorb_rate = 2 // how much item health is removed per tick when absorbing
-	var/absorb_per_health = 3 // how much resources we get per item health
-	var/absorb_completion = 6 // how much resources we get after the item is totally eaten
+	var/absorb_per_health = 1 // how much resources we get per item health
```
stonepillars, line 30: Perhaps a more sane name like `resources_per_health` since it can be confusing when compared with `absorb_rate` and `absorb` later on

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -404,41 +403,58 @@
 
 	var/obj/item/I = absorber.item
 
-	if(I)
-		var/absorb = clamp(src.absorb_rate, 0, I.health)
+	if (!I)
+		return
+
+	var/absorb = min(src.absorb_rate, I.health)
```
stonepillars, line 409: `absorb` -> `absorbed_health`

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -404,41 +403,58 @@
...
+	if (absorber.instant_absorb && !absorber.ignore_amount)
+		boutput(src, "<span class='alert'>[I] is weak enough that it breaks apart instantly!</span>")
+		src.resources += round(src.absorb_per_health * absorb * I.amount)
```
stonepillars, line 412: However improbable, couldn't this lead to an entire stack being absorbed instantly if its current health is less than 1?

FlameArrow57: Yeah, that's why it's included, but it's for resources gained for a single item in the stack being zero. It's to prevent cases like if you were to get zero resources per item in a stack, basically you'd spend the entire time dissolving the item without gaining any resources at all.

stonepillars (approved):
lgtm

PR #273, opened 2022-05-08
Fix bad icon state for spawn-wall sprite

FlameArrow57:
## About the PR 
This PR fixes a bad icon state that had a plus symbol in it instead of a circle for the spawn-wall sprite. Removes an unnecessary state.

## Why's this needed? 
Sprite fix.


stonepillars:
I have no idea what it does and the icon diff does not help so I am going to just merge this and hope nothing explodes

PR #274, opened 2022-05-08
Fix not being able to convert Flock lattices to floors

FlameArrow57:
[BUG]

## About the PR 
This PR just fixes a bug where you couldn't convert Flock lattices to floors.

In doing so, makes it so that the flock convert on harm intent will only do deconstruction, it may not convert tiles anymore.

## Why's this needed? 
Bug fix.


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -889,7 +889,7 @@
 			return
 
 	// CONVERT TURF
-	if(!isturf(target) && !HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING) && !istype(target, /obj/structure/girder))
+	if(!isturf(target) && (!HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING) || istype(target, /obj/lattice/flock)) && !istype(target, /obj/structure/girder))
```
amylizzle, line 892: `obj/lattice/flock` should have the `PROP_ATOM_FLOCK_THING`

stonepillars: I'm pretty sure it does

amylizzle: oh wait it's not'd. NOT has_atom OR istype

PR #275, opened 2022-05-08
Restrict some Flock stuff from being marked for deconstruction or deconstructed

FlameArrow57:
[BUG]

## About the PR 
This PR fixes being able to mark some Flock stuff for deconstruction and deconstruct some stuff that shouldn't be able to be marked or deconstructed.

## Why's this needed? 
Bug fix.

### DIFF
--- a/_std/macros/flock.dm
+++ b/_std/macros/flock.dm
@@ -2,3 +2,4 @@
 #define isfeathertile(x) (istype(x, /turf/simulated/floor/feather) || istype(x, /turf/simulated/wall/auto/feather))
 #define isflock(x) (istype(x, /mob/living/intangible/flock) || istype(x, /mob/living/critter/flock))
 #define isflockstructure(x) (istype(x, /obj/flock_structure))
+#define isflockdeconimmune(x) (istype(target, /obj/flock_structure/ghost) || istype(target, /mob/living/critter/flock) || istype(target, /turf/simulated/floor/feather) || istype(target, /obj/flock_structure/rift) || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
--- a/code/datums/abilities/flock/flockmind.dm
+++ b/code/datums/abilities/flock/flockmind.dm
@@ -459,12 +459,11 @@
 	if(..())
 		return TRUE
 	var/mob/living/intangible/flock/F = holder.owner
-	//special handling for building ghosts
-	if(istype(target,/obj/flock_structure/ghost))
-		//do the tgui window instead
-		//this actually doesn't need bonus behaviour because the cancelbuild is on click, but will need to fix this if we change that in future
-		return TRUE
-	else if(HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING)) //it's a thing we've converted, we can deconstruct it
+	if(HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING))
+		if(istype(target,/obj/flock_structure/ghost))
+			return TRUE // does tgui window for canceling since cancelbuild is on click, will need to fix this if we change it in the future
+		if (istype(target, /mob/living/critter/flock) || istype(target, /turf/simulated/floor/feather) || istype(target, /obj/flock_structure/rift) || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
+			return TRUE
 		F.flock.toggleDeconstructionFlag(target)
 		return FALSE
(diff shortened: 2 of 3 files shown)


code/datums/abilities/flock/flockmind.dm
```diff
@@ -459,12 +459,11 @@
...
-	//special handling for building ghosts
-	if(istype(target,/obj/flock_structure/ghost))
-		//do the tgui window instead
-		//this actually doesn't need bonus behaviour because the cancelbuild is on click, but will need to fix this if we change that in future
-		return TRUE
-	else if(HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING)) //it's a thing we've converted, we can deconstruct it
+	if(HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING))
+		if(istype(target,/obj/flock_structure/ghost))
+			return TRUE // does tgui window for canceling since cancelbuild is on click, will need to fix this if we change it in the future
+		if (istype(target, /mob/living/critter/flock) || istype(target, /turf/simulated/floor/feather) || istype(target, /obj/flock_structure/rift) || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
```
TobleroneSwordfish, line 465:
```suggestion
		if (ismob(target) || istype(target, /turf/simulated/floor/feather) || istype(target, /obj/flock_structure/rift) || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
```
let's just assume we don't want to deconstruct any mobs

TobleroneSwordfish (changes requested):
Might also want to move those typecheck chains into a flock method somewhere for m a i n t a i n a b i l i t y

code/datums/abilities/flock/flockmind.dm
amylizzle, line 465: I think there's a valid argument for being able to deconstruct floor tiles, what with putting a room in vacuum being the only defence against plasma fires.

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -922,9 +922,9 @@
 			else
 				actions.start(new/datum/action/bar/flock_convert(target), user)
 	if(user.a_intent == INTENT_HARM)
-		if(istype(target, /obj/flock_structure/ghost))
-			return
 		if(HAS_ATOM_PROPERTY(target,PROP_ATOM_FLOCK_THING))
+			if(istype(target, /obj/flock_structure/ghost) || istype(target, /turf/simulated/floor/feather) || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
```
amylizzle, line 926:
```suggestion
			if(istype(target, /obj/flock_structure/ghost)  || istype(target, /obj/flock_structure/egg) || istype(target, /obj/flock_structure/relay))
```

code/datums/abilities/flock/flockmind.dm
FlameArrow57: There's nothing currently implemented for deconstructing them so I just put it in. Plus there's possible Flock plating, so think more thought would be maybe needed?

FlameArrow57: I think the atom property should handle that, but can still add if wanted.

stonepillars: I'd prefer some sort of generic way to tag deconstruction blacklist, instead of this copy+pasted if

FlameArrow57: Done, though mob attacking is handled by harm() instead of attack_hand(), even though checked for there in the macro

code/mob/living/critter/flock/flockdrone.dm
FlameArrow57: Above comment

_std/macros/flock.dm
```diff
@@ -1,4 +1,5 @@
 // hi here's some flockdrone BS - cirr
 #define isfeathertile(x) (istype(x, /turf/simulated/floor/feather) || istype(x, /turf/simulated/wall/auto/feather))
 #define isflock(x) (istype(x, /mob/living/intangible/flock) || istype(x, /mob/living/critter/flock))
-#define isflockstructure(x) (istype(x, /obj/flock_structure))
\ No newline at end of file
+#define isflockstructure(x) (istype(x, /obj/flock_structure))
```
stonepillars, line 4: weird

stonepillars (approved):

PR #276, opened 2022-05-08
Drone control icons

TobleroneSwordfish:
[FEATURE]
## About the PR 
sorry pillars
Adds an icon over flockdrones that are being controlled by flockminds or flocktraces, only visible to flock

image   image

## Why's this needed? 
Good to know who is sentient


TobleroneSwordfish:
!merge_upstream

amylizzle:
hows it look with flame's health icons also?

TobleroneSwordfish:
> hows it look with flame's health icons also?

let me merg and find out

TobleroneSwordfish:
image

Health icons don't render on the controller's client however, while the control icons do currently. Not sure which is better.

amylizzle (approved):

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -139,6 +139,8 @@
 	controller = pilot
 	src.client?.color = null // stop being all fucked up and weird aaaagh
 	src.hud?.update_intent()
+	var/image/icon = flock.add_overhead_image('icons/misc/featherzone.dmi', src, pilot.control_icon)
```
amylizzle, line 142: Oh actually maybe this should be in a proc on the flock datum

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -169,6 +171,9 @@
 				controller.mind.key = key
 				controller.mind.current = controller
 				ticker.minds += controller.mind
+		var/image/I = flock.annotations_control_icons[src]
```
amylizzle, line 174: this too

amylizzle (changes requested):
proc up some things

PR #277, opened 2022-05-08
Fix runtimes while checking for disposed busy stuff

stonepillars:
[runtime]
## About the PR 

Fixes `null.disposed` runtimes by switching to `QDELETED`

## Why's this needed? 
Runtimes bad


amylizzle (approved):
gross

PR #286, opened 2022-05-11
Fixed flockdrones being put in clone scanners

### LINKED ISSUE #284: Sticking A Flockdrone Into A Clone Scanner Breaks the Drone
### Labels

[BUG]

### Description

As the title states, putting a flockdrone into a clone scanner completely breaks the drone.

### Steps to reproduce

1. Be a flock drone
2. Click-drag another flock drone into the cloner
3. Release them
4. Notice they have completely shut down, not responding to anything, and not being possessable.

### Expected Behavior

For either them to not break when entering the clone scanner, or not be able to enter the scanner.

Sometimes they will immediately exit the cloner instead of breaking, and sometimes they will enter and break.

TobleroneSwordfish:
[BUG]
## About the PR 
For some reason the clone scanner wasn't using its own helper proc.

## Why's this needed? 
Fixes #284

### DIFF
--- a/code/obj/machinery/computer/cloning.dm
+++ b/code/obj/machinery/computer/cloning.dm
@@ -551,16 +551,7 @@ proc/find_ghost_by_key(var/find_key)
 			return
 
 		var/mob/M = G.affecting
-		M.set_loc(src)
-		src.occupant = M
-		src.icon_state = "scanner_1"
-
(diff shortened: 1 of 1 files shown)


FlameArrow57:
Maybe better upstream?

stonepillars:
Create Upstream. Re-open if there is some **absolutely cursed reason for why it must be like this**.

ZeWaka:
deffo

PR #288, opened 2022-05-11
Flock annotations refactor

TobleroneSwordfish:
[CLEANLINESS]
## About the PR 
Refactors flock annotations to use a single tracking system and switches to using `client_image_group`s to manage image handling.
#253 has been merged into this, with the health icons now using the annotation system and modifying the icon_state instead of re-adding the image every time because I think Flame said they were going to do that anyway.
## Why's this needed? 
A lot less duplicate code, easy to add new annotations and makes them more consistent.


code/datums/flock/flock.dm
```diff
@@ -40,6 +38,8 @@
 	processing_items |= src
 	for(var/DT in childrentypesof(/datum/unlockable_flock_structure))
 		src.unlockableStructures += new DT(src)
+	if (!annotation_imgs)
```
FlameArrow57, line 41: Not necessary considering New() should only be called once?

code/datums/flock/flock.dm
```diff
@@ -228,6 +228,7 @@
 	if(!T)
 		return
 	src.traces -= T
+	hideAnnotations(T)
```
FlameArrow57, line 231: What's reason for putting on removeTrace instead of logout?

FlameArrow57:
Drone control icon also needs to be removed on dormancy, release_control_abrupt(), and disposal.

code/datums/flock/flock.dm
```diff
@@ -18,12 +18,10 @@
 	var/datum/mind/flockmind_mind = null
 	var/list/units = list()
 	var/list/enemies = list()
-	var/list/annotation_viewers = list()
-	var/list/annotations_busy_tiles = list()  // key is atom ref, value is image
-	var/list/annotations_priority_tiles = list()
-	var/list/annotations_deconstruct_targets = list()
-	var/list/annotations_enemies = list()
-	var/list/annotations_control_icons = list()
+	///Associative list of objects to lists of their annotations
```
FlameArrow57, line 21:
```suggestion
	///Associative list of objects to an associative list of their annotation names to images
```
Or something similar

FlameArrow57:
Also, idk, was thinking about how it could be made making so that the controller of a drone can't see their own icons. With how it's implemented, seems pretty simple/trivial to change:

Could just add back in var/list/annotation_viewers, change all cases of
get_image_group(src).add_image(icon) -> old src.addClientImage(icon),

then change get_image_group(src).add_mob(M) ->
src.annotation_viewers += M
if (M.client)
     var/annotation_list
     for (var/thing in src.annotations)
          annotation_list = src.annotations[thing]
          for (var/annotation in annotation_list)
               M.client.images |= annotation_list[annotation]

Similar for removal.

Then for hiding flock face to the controller (similar for health icons when added), after adding the image, could say
if (src.client)
     var/annotations = src.flock.annotations[src]
     src.client.images -= annotations["flockface"]

code/datums/flock/flock.dm
TobleroneSwordfish: `annotation_imgs` is static, so it's shared between all instances

code/datums/flock/flock.dm
TobleroneSwordfish: By the time `Logout` is called the client has already been transferred out of the mob, and you need the client to remove the image. I'm pretty sure the calls to `hideAnnotations` in logout never worked.

TobleroneSwordfish:
> Drone control icon also needs to be removed on dormancy, release_control_abrupt(), death, and disposal. Might be good to just make a new proc for toggling the control icon?

It now is removed on all of those, I've removed the multiple copy pasted implementations of `release_control` to help fix this and because scope creeeep yay. I haven't added a proc for toggling the control icon since making specific procs for each icon is what we are trying to factor out here.

FlameArrow57:
With the release_control changes, it kind of loses the special message inside of dormantize(). Also, the message you are supposed to be given for deletion of the drone when in the drone isn't given.

TobleroneSwordfish:
alright, dormantizing a drone you control now looks like this 
image
and deleting looks like this
image
is that okay?

stonepillars (commented):
mandatory nitpick

stonepillars (approved):
lgtm

code/datums/flock/flock.dm
```diff
@@ -485,14 +499,12 @@
 		hideAnnotations(src.flockmind)
 	for(var/mob/M in src.units)
 		hideAnnotations(M)
+	qdel(get_image_group(src))
 	all_owned_tiles = null
 	busy_tiles = null
 	priority_tiles = null
 	units = null
 	enemies = null
```
FlameArrow57, line 507: Clear annotations var here?

code/datums/flock/flock.dm
FlameArrow57: Idk, I think this could still be put in logout? I think intention was just to remove client images while there's no client there, as showAnnotations() and hideAnnotations() is also done for flockdrones.

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -199,43 +178,37 @@
...
+		flock.removeAnnotation(src, FLOCK_ANNOTATION_FLOCKTRACE_CONTROL)
+	if (give_alerts)
+		boutput(controller, "<span class='flocksay'><b>\[SYSTEM: Control of drone [src.real_name] ended abruptly.\]</b></span>")
 	controller = null
 	src.update_health_icon()
 
 /mob/living/critter/flock/drone/dormantize()
 	src.icon_state = "drone-dormant"
 	src.remove_simple_light("drone_light")
+	emote("beep")
```
FlameArrow57, line 193: Was originally meant to be quiet (no beep)

FlameArrow57:
Able to just make a third proc to do stuff that's done in release_control and release_control_abrupt? The second was originally intended to be its own function, so feels off having release_control call it when they're two different things

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -139,46 +139,30 @@
...
-/mob/living/critter/flock/drone/proc/release_control_abrupt()
+	_drop_control()
+	if (give_alerts)
+		flock_speak(null, "Control of drone [src.real_name] surrended.", src.flock)
+
+///Release control of a drone without blocking calls, so it can be used in ghostize etc.
+/mob/living/critter/flock/drone/proc/release_control_abrupt(give_alerts = TRUE)
+	_drop_control()
+	if (give_alerts)
+		boutput(controller, "<span class='flocksay'><b>\[SYSTEM: Control of drone [src.real_name] ended abruptly.\]</b></span>")
```
FlameArrow57, line 162:
```suggestion
	if (give_alerts)
		boutput(controller, "<span class='flocksay'><b>\[SYSTEM: Control of drone [src.real_name] ended abruptly.\]</b></span>")
	_drop_control()
```
Don't think message will actually be sent if drop control is done first?

code/datums/flock/flock.dm
```diff
@@ -313,48 +313,95 @@
...
+	.[FLOCK_ANNOTATION_FLOCKMIND_CONTROL] = flock_face
+
+	var/image/trace_face = image('icons/misc/featherzone.dmi', icon_state = "flocktrace_face")
+	trace_face.blend_mode = BLEND_ADD
+	trace_face.plane = PLANE_ABOVE_LIGHTING
+	trace_face.appearance_flags = RESET_COLOR | RESET_ALPHA | RESET_TRANSFORM
+	trace_face.pixel_y = 16
+	.[FLOCK_ANNOTATION_FLOCKTRACE_CONTROL] = trace_face
+
+	var/image/health = image('icons/misc/featherzone.dmi', src, "hp-100")
```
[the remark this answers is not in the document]
FlameArrow57: Didn't seem like was fixed completely? I mean idk if there's actually a problem, but seems like something was started

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -139,46 +139,30 @@
...
-		controller = null
-		src.update_health_icon()
-
-/mob/living/critter/flock/drone/proc/release_control_abrupt()
+	_drop_control()
+	if (give_alerts)
+		flock_speak(null, "Control of drone [src.real_name] surrended.", src.flock)
+
+///Release control of a drone without blocking calls, so it can be used in ghostize etc.
+/mob/living/critter/flock/drone/proc/release_control_abrupt(give_alerts = TRUE)
```
FlameArrow57, line 159:
```suggestion
/mob/living/critter/flock/drone/proc/release_control_abrupt()
```

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -210,32 +197,20 @@
 	if (!src.flock)
 		..()
 		return
-
 	src.flock.hideAnnotations(src)
-
+	//hold a ref to the controller so we can put them back with the flock
+	var/mob/living/intangible/flock/old_controller = src.controller
 	if (src.controller)
+		src.release_control_abrupt(FALSE)
```
FlameArrow57, line 204:
```suggestion
		src.release_control(FALSE)
```

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: actually it should just use `flock_speak`, no idea why it wasn't

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -139,46 +139,30 @@
...
-		if (give_alerts)
-			flock_speak(null, "Control of drone [src.real_name] surrended.", src.flock)
-		// clear refs
-		controller = null
-		src.update_health_icon()
-
-/mob/living/critter/flock/drone/proc/release_control_abrupt()
+	_drop_control()
+	if (give_alerts)
+		flock_speak(null, "Control of drone [src.real_name] surrended.", src.flock)
```
FlameArrow57, line 156: Not necessary, but maybe could just put _drop_control() first and then have one give_alerts section?

code/datums/flock/flock.dm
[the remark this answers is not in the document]
TobleroneSwordfish: oh there's a `src` there doh

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -210,32 +197,20 @@
...
 		..()
 		return
-
 	src.flock.hideAnnotations(src)
-
+	//hold a ref to the controller so we can put them back with the flock
+	var/mob/living/intangible/flock/old_controller = src.controller
 	if (src.controller)
+		src.release_control_abrupt(FALSE)
+	if (old_controller)
```
FlameArrow57, line 205:
```suggestion
```

FlameArrow57:
Could also remove showAnnotations() and hideAnnotations() from flockdrone login and logout since it's done for flocktraces?

code/mob/living/critter/flock/flockdrone.dm
FlameArrow57: The reason why was because every place release_control() was being used before was on deletion or ghostizing, in which case the death message for flock drones is sent already, so no need to tell the flock, kind of like how controlling a drone doesn't tell flock

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: you said you wanted dormantize to not beep, `release_control` beeps

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: that would runtime if you call dormantize on a drone with no controller

FlameArrow57: Yeah, but if there's old_controller, there's also a controller, since old_controller was assigned to controller?

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: but why? the other release_control has this param and it's used in code

FlameArrow57: On changing release_control_abrupt() in dormantize() to release_control(), all places would have TRUE passed in, and then unnecessary

TobleroneSwordfish:
> Could also remove showAnnotations() and hideAnnotations() from flockdrone login and logout, and showAnnotations() for intangible flock login since it's done for flocktraces?

Flockdrones don't define logout, and removing those showAnnotations calls causes the annotation to not be shown.

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: correct, but we don't know that there was a controller

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: sure

code/datums/flock/flock.dm
```diff
@@ -505,22 +518,12 @@
...
 	src.busy_tiles[name] = T
-
-	var/image/I = image('icons/misc/featherzone.dmi', T.RL_MulOverlay ? T.RL_MulOverlay : T, "frontier")
-	I.appearance_flags = RESET_ALPHA | RESET_COLOR
-	I.alpha = 80
-	I.plane = PLANE_ABOVE_LIGHTING
-	I.mouse_opacity = FALSE
-	src.annotations_busy_tiles[T] = I
-	src.addClientImage(I)
+	addAnnotation(T, "reserve")
```
stonepillars, line 521:
```suggestion
	addAnnotation(T, FLOCK_ANNOTATION_RESERVED)
```
don't you mean?

TobleroneSwordfish: yes I do mean

code/datums/flock/flock.dm
```diff
@@ -505,22 +518,12 @@
...
+	addAnnotation(T, FLOCK_ANNOTATION_RESERVED)
 
 /datum/flock/proc/unreserveTurf(var/name)
 	var/turf/simulated/T = src.busy_tiles[name]
 	src.busy_tiles -= name
-
-	var/image/I = src.annotations_busy_tiles[T]
-	src.annotations_busy_tiles -= T
-	src.removeClientImage(I)
+	removeAnnotation(T, "reserve")
```
stonepillars, line 526:
```suggestion
	removeAnnotation(T, FLOCK_ANNOTATION_RESERVED)
```

PR #289, opened 2022-05-12
Efficient AI please

amylizzle:
## About the PR 
Gonna do some minor tweaks to flock AI to make it more efficient.

- [x] Make move only try to calculate the path once before failing.
- [x] Make move have a shorter maximum path unless you're rallying.
- [x] Check out preconditions, see if I can't make them prevent doing `get_path_to` with clever checks.
- [x] Look at trimming down the calls to `isEnemy()`, which has a ludicrous number of calls.

### DIFF
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -763,33 +763,21 @@ butcher
 	var/mob/living/critter/flock/drone/F = holder.owner
 	if(!F?.flock)
 		return
-	for(var/atom/T in view(target_range, holder.owner))
-		//handle vehicles first
-		if (isvehicle(T))
-			var/vehicle_is_enemy = F.flock.isEnemy(T) //if someone gets in an enemy pod, attack
-			var/enemy_found = FALSE
-			for (var/mob/occupant in T)
-				if (!F.flock.isEnemy(occupant) && !vehicle_is_enemy)
+	for(var/atom/T in F.flock.enemies)
+		if(IN_RANGE(T,holder.owner,target_range))
+			F.flock.updateEnemy(T)
+			if(isliving(T))
+				var/mob/living/M = T
+				if(is_incapacitated(M))
 					continue
-				F.flock.updateEnemy(occupant)
-				enemy_found = TRUE
-				break
-			if (enemy_found) //bad guy in pod
-				. += T
-			//continue regardless of whether we find an enemy or not, since we don't want to attack empty pods
-			continue
-		if(!F.flock.isEnemy(T))
-			continue
-		if(isliving(T))
-			var/mob/living/M = T
-			if(is_incapacitated(M))
+			if(istype(T.loc.type, /obj/flock_structure/cage))
 				continue
-		// mob is a valid target, check if they're not already in a cage
-		if(!istype(T.loc.type, /obj/flock_structure/cage))
-			// if we can get a valid path to the target, include it for consideration
 			. += T
-		F.flock.updateEnemy(T)
-
+		else if (isvehicle(T.loc))
+			if(IN_RANGE(T.loc, holder.owner, target_range))
+				F.flock.updateEnemy(T)
+				F.flock.updateEnemy(T.loc)
+				. += T.loc
 
 ///////////////////////////////////////////////////////////////////////////////////////////////////////////
 // FLOCKDRONE-SPECIFIC CAPTURE TASK
@@ -834,11 +822,11 @@ butcher
 	. = list()
 	var/mob/living/critter/flock/drone/F = holder.owner
 	if(F?.flock)
-		for(var/atom/T in view(max_dist, holder.owner))
-			if (valid_target(T))
-				// if we can get a valid path to the target, include it for consideration
-				. += T
-				F.flock.updateEnemy(T)
+		for(var/atom/T in F.flock.enemies)
+			if(IN_RANGE(T,holder.owner,max_dist))
+				if (valid_target(T))
+					. += T
+					F.flock.updateEnemy(T)
 	. = get_path_to(holder.owner, ., max_dist*2, 1)
 
 /datum/aiTask/succeedable/capture
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -78,9 +78,9 @@ butcher
 
 	New()
 		..()
-		var/datum/aiTask/succeedable/move/tmp = subtasks[subtask_index]
-		if(istype(tmp))
-			tmp.max_path_dist = 300
+		var/datum/aiTask/succeedable/move/movesubtask = subtasks[subtask_index]
+		if(istype(movesubtask))
+			movesubtask.max_path_dist = 300
 
 	on_tick()
 		if (!holder.target)
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -776,7 +776,7 @@ butcher
 	if(!F?.flock)
 		return
 	for(var/atom/T in F.flock.enemies)
-		if(IN_RANGE(T,holder.owner,target_range))
+		if(T in view(holder.owner,target_range))
 			F.flock.updateEnemy(T)
 			if(isliving(T))
 				var/mob/living/M = T
@@ -786,7 +786,7 @@ butcher
 				continue
 			. += T
 		else if (isvehicle(T.loc))
-			if(IN_RANGE(T.loc, holder.owner, target_range))
+			if(T.loc in view(holder.owner, target_range))
 				F.flock.updateEnemy(T)
 				F.flock.updateEnemy(T.loc)
 				. += T.loc
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -1155,6 +1155,12 @@ butcher
 
 /////// Targetable AI tasks, instead of looking for targets around them they just override with their own target var
 /datum/aiTask/sequence/goalbased/build/targetable
+	New()
+		..()
+		var/datum/aiTask/succeedable/move/movesubtask = subtasks[subtask_index]
+		if(istype(movesubtask))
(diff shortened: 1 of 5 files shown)


code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -76,6 +76,12 @@ butcher
 	can_be_adjacent_to_target = FALSE
 	max_dist = 0
 
+	New()
+		..()
+		var/datum/aiTask/succeedable/move/tmp = subtasks[subtask_index]
```
TobleroneSwordfish, line 81: plz don't call variables `tmp` :(

TobleroneSwordfish (changes requested):
The other targetable tasks should probably also get a range increase

code/mob/living/critter/ai/flock/flocktasks.dm
amylizzle: but tmp is their function. they are tmp. (will fix)

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -713,9 +714,20 @@ butcher
...
+	var/mob/living/critter/flock/drone/F = holder.owner
+	if(length(F.flock?.enemies))
+		var/datum/handHolder/HH = F.hands[3]
+		var/datum/limb/gun/stunner = HH?.limb
+		if(istype(stunner) && !stunner.is_on_cooldown())
+			return TRUE
 
 /datum/aiTask/timed/targeted/flockdrone_shoot/evaluate()
-	return weight * score_target(get_best_target(get_targets()))
+	if(src.precondition())
```
TobleroneSwordfish, line 727: Surely this check is already done by the AI holder?

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -713,9 +714,20 @@ butcher
 	var/run_range = 3
 	var/list/dummy_params = list("icon-x" = 16, "icon-y" = 16)
 
+/datum/aiTask/timed/targeted/flockdrone_shoot/proc/precondition()
+	//if we know there are enemies, and gun is charged and ready
+	var/mob/living/critter/flock/drone/F = holder.owner
+	if(length(F.flock?.enemies))
+		var/datum/handHolder/HH = F.hands[3]
+		var/datum/limb/gun/stunner = HH?.limb
+		if(istype(stunner) && !stunner.is_on_cooldown())
```
TobleroneSwordfish, line 723: This will mean a drone with its stunner on cooldown will just calmly revert to harvesting or something, rather than reacting like its in combat. Maybe a better change would be to make the drone dodge if their stunner is on cooldown during a shoot task, so they still react as if they're in combat and try to dodge until they can shoot back.

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -757,33 +775,21 @@ butcher
...
 		return
-	for(var/atom/T in view(target_range, holder.owner))
-		//handle vehicles first
-		if (isvehicle(T))
-			var/vehicle_is_enemy = F.flock.isEnemy(T) //if someone gets in an enemy pod, attack
-			var/enemy_found = FALSE
-			for (var/mob/occupant in T)
-				if (!F.flock.isEnemy(occupant) && !vehicle_is_enemy)
+	for(var/atom/T in F.flock.enemies)
+		if(IN_RANGE(T,holder.owner,target_range))
```
TobleroneSwordfish, line 779: Won't this result in drones trying to shoot people through walls? You could switch to `if (T in holder.owner.view)` which would probably still be faster than iterating over view.

code/procs/helpers.dm
```diff
@@ -266,6 +266,15 @@ proc/get_angle(atom/a, atom/b)
 			return 1
 	return 0
 
+//is_blocked_turf for flock
+/proc/flock_is_blocked_turf(var/turf/T)
+	if (!T) return 0
```
TobleroneSwordfish, line 271: Switch to TRUE/FALSE to make Flame 200% less sad

TobleroneSwordfish (changes requested):
fast ai good

code/mob/living/critter/ai/flock/flocktasks.dm
amylizzle: it is not. Precondition is defined for sequential aitasks, but not for timed for some bizarre reason.

code/mob/living/critter/ai/flock/flocktasks.dm
amylizzle: My reasoning for this was that it's totally possible to essentially "combat lock" a bunch of drones by having them attack you but not be able to get you (see: an enemy pod parked outside a locked airlock or something). I figured if we make them do other stuff between gun charges, it's less likely to be a cheesey strategy because they'll be able to convert doors/walls and do other stuff between taking pot shots.

code/mob/living/critter/ai/flock/flocktasks.dm
amylizzle: yeah, that would probably be better.

code/mob/living/critter/ai/flock/flocktasks.dm
TobleroneSwordfish: image

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -1197,6 +1221,7 @@ butcher
 		holder.target = get_turf(src.target)
 
 /datum/aiTask/timed/targeted/flockdrone_shoot/targetable
+
```
stonepillars, line 1224: ?

amylizzle: I was gonna put increased targetting range in this one two and then I realised that it wouldn't work because shoot does it differently, so I removed it - accidental white-space was the result.

amylizzle: I was gonna put increased targetting range in this one two and then I realised that it wouldn't work because shoot does it differently, so I removed it - accidental white-space was the result.

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -757,33 +775,21 @@ butcher
...
-		if(!istype(T.loc.type, /obj/flock_structure/cage))
-			// if we can get a valid path to the target, include it for consideration
 			. += T
-		F.flock.updateEnemy(T)
-
+		else if (isvehicle(T.loc))
+			if(T.loc in view(holder.owner, target_range))
+				F.flock.updateEnemy(T)
+				F.flock.updateEnemy(T.loc)
+				. += T.loc
```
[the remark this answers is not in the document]
amylizzle: wat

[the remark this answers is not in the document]
stonepillars: i cant read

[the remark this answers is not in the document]
amylizzle: resolved in discord

stonepillars (approved):
lgtm

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -757,33 +775,21 @@ butcher
...
-			if (enemy_found) //bad guy in pod
-				. += T
-			//continue regardless of whether we find an enemy or not, since we don't want to attack empty pods
-			continue
-		if(!F.flock.isEnemy(T))
-			continue
-		if(isliving(T))
-			var/mob/living/M = T
-			if(is_incapacitated(M))
+			if(istype(T.loc, /obj/flock_structure/cage))
```
FlameArrow57, line 785: For T.loc, cage is checked here, but vehicle is checked for outside?

amylizzle: vehicles can't be in cages

code/obj/flock/cage.dm
```diff
@@ -195,6 +195,7 @@
 				eating_occupant = 0
 				playsound(src, "sound/weapons/nano-blade-1.ogg", 50, 1)
 				if(occupant)
+					src?.flock?.updateEnemy(occupant)
```
FlameArrow57, line 198: I think this should be moved out, it means that the enemy will only be updated if it has something that can be eaten?

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -757,32 +775,22 @@ butcher
...
+			if(T.loc in view(holder.owner, target_range))
+				F.flock.updateEnemy(T)
+				F.flock.updateEnemy(T.loc)
+				. += T.loc
+		else if(T in view(holder.owner,target_range))
+			F.flock.updateEnemy(T)
+			if(isliving(T))
+				var/mob/living/M = T
+				if(is_incapacitated(M))
+					continue
```
stonepillars, line 792: Why doesn't this call `isEnemy` anywhere?

amylizzle: Because we're iterating over the list of enemies

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -76,6 +76,12 @@ butcher
 	can_be_adjacent_to_target = FALSE
 	max_dist = 0
 
+	New()
+		..()
+		var/datum/aiTask/succeedable/move/movesubtask = subtasks[subtask_index]
+		if(istype(movesubtask))
+			movesubtask.max_path_dist = 300
```
stonepillars, line 83: Is there some way to avoid copy+pasting this?

TobleroneSwordfish: I don't think so, the reason it's copy pasted is because the targetable tasks are horrible hacks that don't actually share inheritance. I guess you could define a proc to set up an AItask as targetable and then call that from the new of all the targetables, but that might just be even more cursed.

amylizzle: I guess I could add a flag to the task for long range moves or not, but tbh I think that might be better to do in another pr where I also slightly refractor timed tasks to use it too.

PR #290, opened 2022-05-13
Properly handle Flockmind and Flocktrace suicides

### LINKED ISSUE #155: Suiciding as a drone causes a cascade of runtimes
### Labels

[BUG]

### Description

Suicide as a drone, runtime

### Steps to reproduce

1. Take control of drone
2. type suicide command
3. you are kicked out the drone
4. the TGUI suicide confirm is still there
5. click yes
6. runtime

### Expected Behavior

Honestly not sure - should drones be able to suicide?

Pretty sure the runtimes were because the busy_tiles list was null, we should probably initialise it. But also, given that I was kicked out of the drone and the drone then went about its business, it seems like the suicide command was caught and then passed somewhere that did normal suicide behaviour which flock predates.

FlameArrow57:
[BUG]

## About the PR 
This PR adds in proper handling for Flockmind and Flocktrace suicides.

Flockminds may not suicide if the relay is up and they have no active traces available to promote. Otherwise, the Flockmind and Flocktraces may suicide at any time.

For the Flockmind, when suiciding with an active trace available, must choose one to promote to the new Flockmind before suiciding.

Two bugs are known:
1. The Flockmind when suiciding, if there is an active trace in a Flockdrone, in admin view the Flockdrone will have the antag V symbol on it even when the Flocktrace exits. Was thinking maybe this was an issue out of scope of the PR.
2. The "You cease to exist abruptly" message for Flocktraces suiciding is not given for when you suicide in a drone. Did not why or how to fix this.

## Why's this needed? 
Bug fixes, fixes #155.


code/datums/flock/flock.dm
```diff
@@ -478,8 +492,12 @@
...
 	if(src.flockmind)
 		hideAnnotations(src.flockmind)
-	for(var/mob/M in src.units)
-		hideAnnotations(M)
+	for(var/mob/living/intangible/flock/trace/T as anything in src.traces)
+		T.death()
+	for(var/mob/living/critter/flock/F as anything in src.units)
+		F.dormantize()
+	for(var/obj/flock_structure/S as anything in src.structures)
+		S.flock = null
```
amylizzle, line 500:
```suggestion
		src.removeStructure(S)
```
and probably add `S.flock = null` to that function. It's probably an oversight that we hadn't done that already.

code/mob/suicide.dm
```diff
@@ -180,14 +180,35 @@
...
-		C.unlock_medal("Damned", 1)
-
-	return TRUE
+	if ((locate(/obj/flock_structure/relay) in src.flock.structures) && istype(src.controller, /mob/living/intangible/flock/flockmind) && !length(src.flock.getActiveTraces()))
+		boutput(src, "<span class='alert'>You can't abandon your Flock with the Relay active!</span>")
+		return
+	if (tgui_alert(src, "Are you sure you want to commit suicide?", "Confirm Suicide", list("Yes", "No")) == "Yes")
+		var/mob/living/intangible/flock/C = src.controller
+		src.release_control()
+		if (istype(C, /mob/living/intangible/flock/flockmind))
```
amylizzle, line 190: Add a `do_suicide` stub proc to `/mob/living/intangible/flock` and override it in flockmind and trace, you can skip this check and just do `C.do_suicide()` instead

code/mob/suicide.dm
```diff
@@ -180,14 +180,35 @@
...
+		else
+			var/mob/living/intangible/flock/trace/T = C
+			T.do_suicide(TRUE)
+
+/mob/living/intangible/flock/flockmind/do_suicide(skip_prompt)
+	if ((locate(/obj/flock_structure/relay) in src.flock.structures) && !length(src.flock.getActiveTraces()))
+		boutput(src, "<span class='alert'>You can't abandon your Flock with the Relay active!</span>")
+		return
+	if (skip_prompt || tgui_alert(src, "Are you sure you want to commit suicide?", "Confirm Suicide", list("Yes", "No")) == "Yes")
+		var/result = src.getTraceToPromote()
```
amylizzle, line 202: What does this return if the flockmind that is suiciding just closes the game or something instead of picking from the list?

code/obj/flock/structure/rift.dm
```diff
@@ -45,7 +43,10 @@
 					candidate_turfs -= S
 					continue
 				if(prob(25))
-					src.flock.claimTurf(flock_convert_turf(S))
+					if (src.flock)
```
amylizzle, line 46: what's this for?

amylizzle (changes requested):

code/obj/flock/structure/rift.dm
FlameArrow57: If you suicide while you have a rift

amylizzle: if you suicide with a rift, it should kill the rift.

code/mob/suicide.dm
FlameArrow57: Idk, they don't really have anything common to them except for a single tgui prompt, would just end up doing an if statement for type and doing what was done in each function.

code/mob/suicide.dm
FlameArrow57: Nothing happens when closing the game, so would think it just cancels the prompt

code/mob/suicide.dm
amylizzle: That's fine, by stub proc I just mean an empty proc definition. All intangible flock things are going to need it, and it's better for future code that it's defined in the parent class. I think there's even a flag you can use to force all child classes to implement it.

code/mob/suicide.dm
amylizzle: Maybe pick a trace at random if the prompt is cancelled?

FlameArrow57: Idk, would say not necessary, since still some time to back out if you want I suppose. If done to prevent just quitting the game, will need some afk timer check in the future for quitting without suiciding, so could probably just use that instead?

amylizzle: hmm, maybe you're right. I guess leave this for now, and if it actually becomes an issue in the future, we can fix it then.

code/mob/suicide.dm
```diff
@@ -180,14 +180,35 @@
...
+		if (istype(result, /mob/living/intangible/flock/trace))
+			src.unlock_medal("Damned", 1)
+			var/mob/living/intangible/flock/trace/T = result
+			T.promoteToFlockmind(TRUE)
+		else if (result == -1)
+			src.unlock_medal("Damned", 1)
+			src.death(suicide = TRUE)
```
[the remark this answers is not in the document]
FlameArrow57: There's the option to cancel the prompt, in which case you wouldn't suicide and couldn't get the medal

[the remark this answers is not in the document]
stonepillars: Yes but this if is only ever reached if they say yes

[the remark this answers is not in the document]
FlameArrow57: I was meaning for if you press cancel while selecting a trace

code/obj/flock/structure/rift.dm
FlameArrow57: Unless there's a specific reason for deleting the rift, I think it could be done either way?

amylizzle: I say delete the rift because without a flockmind, there isn't really a flock to spawn. I guess you could do something like triggering the partition ability and asking for a ghost to take on the role when the egg pops.

code/mob/suicide.dm
FlameArrow57: Hm?

code/mob/suicide.dm
FlameArrow57: Both do_suicide() functions have a parameter passed in, but if having a stub proc, it wouldn't have one, or if it does it would be unused?

amylizzle: You can define it and not use it, no problem :)

code/mob/suicide.dm
TobleroneSwordfish: I'm saying that cancel should just close the prompt, not do something else.

code/obj/flock/structure/rift.dm
amylizzle: resolved, by medass

stonepillars: image

stonepillars:
Ignore that, meant to mention another PR. This PR has a tangentially related issue: #306

PR #292, opened 2022-05-13
Add flock parameter to Flock mass conversion procs

FlameArrow57:
[BUG]

## About the PR 
This PR just adds a flock parameter to Flock mass conversion procs, also making the relay use it.

## Why's this needed? 
Fixes a bug where stuff converted to Flock by the relay wasn't registering its Flock.

Would be nice to have for the other proc if wanted to be used.

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -737,16 +737,16 @@
 
 	return T
 
-/proc/mass_flock_convert_turf(var/turf/T)
+/proc/mass_flock_convert_turf(var/turf/T, datum/flock/F)
 	// a terrible idea
 	if(!T)
 		T = get_turf(usr)
 	if(!T)
 		return // not sure if this can happen, so it will
 
-	flock_spiral_conversion(T)
+	flock_spiral_conversion(T, F)
 
-/proc/radial_flock_conversion(var/atom/movable/source, var/max_radius=20)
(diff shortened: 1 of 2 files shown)


amylizzle (approved):

code/datums/flock/flock.dm
```diff
@@ -755,7 +755,10 @@
 		LAGCHECK(LAG_LOW)
 		for(var/turf/tile in turfs)
 			if(istype(tile, /turf/simulated) && !isfeathertile(tile))
-				flock_convert_turf(tile)
+				if (F)
+					F.claimTurf(flock_convert_turf(tile))
+				else
+					flock_convert_turf(tile)
```
stonepillars, line 761:
```suggestion
				flock_convert_turf(tile)
				if (F)
					F.claimTurf(tile)
```

This might just be pedantic, but. Feel free to just resolve.

TobleroneSwordfish: actually not sure that'd work, since I think `flock_convert_turf` deletes the old turf and returns the new one

TobleroneSwordfish: actually I don't think that'd work, since `flock_convert_turf` deletes the old turf and returns the new one

FlameArrow57: It does, if you search in the code for other replacements similar, there's cases of this. Idk why it works.

stonepillars: turfs are wack

PR #293, opened 2022-05-13
Fix some dormancy bugs

FlameArrow57:
[BUG]

## About the PR 
This PR just fixes some bugs with dormantizing.

Fixes being in the last Flockdrone dormantized not moving you back to the station, ejecting from a Flockdrone on another z level not moving you back to the station, Flockdrones and Flockbits being dormantized in closets, Flockdrones and Flockbits continuing actions when dormantized, items being absorbed in dormantized Flockdrones, and being able to open the context menu of dormantized Flockdrones.

Makes it so that Flockdrones and Flockbits are no longer dormantized on the null z level in the process.

## Why's this needed? 
Bug fixes.


code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -185,7 +194,16 @@
 		return
 	if (src.floorrunning)
 		src.end_floorrunning(TRUE)
-	controller.set_loc(get_turf(src))
+	if (src.z == Z_LEVEL_STATION)
```
TobleroneSwordfish, line 197: copy pasta, please make it a proc

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -156,7 +156,16 @@
...
 		// move controller out
-		controller.set_loc(get_turf(src))
+		if (src.z == Z_LEVEL_STATION)
+			controller.set_loc(get_turf(src))
+		else
+			if (src.flock?.getComplexDroneCount() > 1)
+				for (var/mob/living/critter/flock/drone/F in src.flock.units)
+					if (istype(F) && F != src)
+						src.controller.set_loc(get_turf(F))
+						break
```
stonepillars, line 166: Why this horrible loop? Why not just send them to latejoin or blobstart whenever they exit something on another zlevel? That's what ghosts do when they're on z2 and z4

FlameArrow57: I moved it into a proc, so only one case of it appearing in code. Other than that, I don't see the problem with it? It just makes it a little faster.

stonepillars: It's faster to just not go over
Everything until you get a drone. I'd prefer `pick` if you wanted similar behaviour, or just sending them to a landmark like blob start

stonepillars: Then just jump to a landmark

PR #297, opened 2022-05-14
Makes pinging turfs a little less weird

TobleroneSwordfish:
[two line PR]
## About the PR 
Makes pinging a turf apply an alpha mask to the ping overlay so it doesn't obscure the atoms on the turf. It's limited to turfs because it looks very weird on other things.
image image
Pinging the turf vs pinging a thing on the turf.

## Why's this needed? 
Looks better


stonepillars (approved):
lgtm

PR #298, opened 2022-05-15
Fixes very threatening walls

### LINKED ISSUE #295: Reflected shots cause walls to be threatening
### Labels

[BUG]

### Description

Walls are marked by the protection component

### Steps to reproduce

1. Convert walls in a room
2. Have a drone shoot something and miss
3. Shot bounces off of the flock walls
4. Hits another flock thing
5. 
image

### Expected Behavior

Reflected shots probably shouldn't count for threats? or something

_No response_

TobleroneSwordfish:
[BUG]
## About the PR 
Fixes #295 and several adjacent bugs in flockdrone harm reporting code, including a condition which checked a proc which always returned false and the protection component mixing up thrown and projectile flags.

## Why's this needed? 
Walls are now far less threatening.

### DIFF
--- a/code/datums/flock/flockprotection.dm
+++ b/code/datums/flock/flockprotection.dm
@@ -79,5 +79,5 @@
 
 /// Protect against someone shooting the parent.
 /datum/component/flock_protection/proc/handle_hitby_proj(atom/source, obj/projectile/P)
-	var/attacker = P.shooter
+	var/attacker = P.mob_shooter
 	return attacker && src.report_thrown && SEND_SIGNAL(source, COMSIG_FLOCK_ATTACK, attacker, FALSE)
--- a/code/datums/flock/flockprotection.dm
+++ b/code/datums/flock/flockprotection.dm
@@ -81,5 +81,5 @@
 /datum/component/flock_protection/proc/handle_hitby_proj(atom/source, obj/projectile/P)
 	var/attacker = P.shooter
 	if (!(ismob(attacker) || iscritter(attacker) || isvehicle(attacker)))
-		attacker = P.mob_shooter
+		attacker = P.mob_shooter //shooter is updated on reflection, so we fall back to mob_shooter if it turns out to be a wall or something
 	return attacker && src.report_proj && SEND_SIGNAL(source, COMSIG_FLOCK_ATTACK, attacker, FALSE)
--- a/code/datums/flock/flockprotection.dm
+++ b/code/datums/flock/flockprotection.dm
@@ -80,6 +80,6 @@
 /// Protect against someone shooting the parent.
 /datum/component/flock_protection/proc/handle_hitby_proj(atom/source, obj/projectile/P)
 	var/attacker = P.shooter
-	if (!ismob(P.shooter))
+	if (!(ismob(attacker) || iscritter(attacker) || isvehicle(attacker)))
 		attacker = P.mob_shooter
 	return attacker && src.report_proj && SEND_SIGNAL(source, COMSIG_FLOCK_ATTACK, attacker, FALSE)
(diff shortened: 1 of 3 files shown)


FlameArrow57:
See: https://github.com/stonepillars/goon-flock/pull/86

If I remember right, it was only working for point blanking or something, but don't remember exactly.

Also, idk, I think ideally reflected projectiles should be retaining their original shooter.

TobleroneSwordfish:
Huh weird, okay let me test it more

code/datums/flock/flockprotection.dm
```diff
@@ -80,4 +80,6 @@
 /// Protect against someone shooting the parent.
 /datum/component/flock_protection/proc/handle_hitby_proj(atom/source, obj/projectile/P)
 	var/attacker = P.shooter
-	return attacker && src.report_thrown && SEND_SIGNAL(source, COMSIG_FLOCK_ATTACK, attacker, FALSE)
+	if (!ismob(P.shooter))
```
amylizzle, line 83: obj critters and vehicles?

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -592,10 +594,12 @@
...
 	if(floorrunning)
 		return FALSE
-	if (..())
-		var/attacker = P.shooter
-		if(attacker)
-			src.harmedBy(attacker)
+	if (!..())
+		return
+	var/attacker = P.shooter
+	if(!ismob(attacker))
```
amylizzle, line 600: ditto

amylizzle (changes requested):

code/mob/living/critter/flockcritter_parent.dm
```diff
@@ -105,6 +105,7 @@
 		src.visible_message("<span class='notice'>[src] harmlessly absorbs [P].</span>")
 		return FALSE
 	..()
+	return TRUE
```
FlameArrow57, line 108:
```suggestion
	return ..()
```

TobleroneSwordfish: I think it makes sense for flockdrones to still be offended by being attacked even if the shot reflects off them? (don't actually think this can happen currently, but someone said something about making flockdrones reflect phaser shots)

FlameArrow57: They can't right now, but reflected shots when testing were still hitting if I remember right

stonepillars (approved):

code/datums/flock/flockprotection.dm
```diff
@@ -80,4 +80,6 @@
 /// Protect against someone shooting the parent.
 /datum/component/flock_protection/proc/handle_hitby_proj(atom/source, obj/projectile/P)
 	var/attacker = P.shooter
-	return attacker && src.report_thrown && SEND_SIGNAL(source, COMSIG_FLOCK_ATTACK, attacker, FALSE)
+	if (!(ismob(attacker) || iscritter(attacker) || isvehicle(attacker)))
+		attacker = P.mob_shooter
```
FlameArrow57, line 84:
```suggestion
		attacker = P.mob_shooter // accounts for mob-fired reflected projectiles and things like turrets
```
Or something similar, and same for below?

FlameArrow57:
Bug for future reference (very small): Reflected projectiles from non-mobs won't be reported (I don't think)

TobleroneSwordfish:
> Bug for future reference (very small): Reflected projectiles from non-mobs won't be reported (I don't think)

Given we can't mark a turret as an enemy it's not really an issue.

FlameArrow57:
It will happen for pods though I think

TobleroneSwordfish:
> True, was thinking more of pods

pods are vehicles, so should be handled fine

PR #299, opened 2022-05-17
Minor balance change

amylizzle:
## About the PR 
Prevents building (converting walls/floors) before 10 drones
Slightly higher absorbtion rate, still lower than original.

## Why's this needed? 
Drones weren't reproducing and until all the walls around them were converted, and it was really slowing down the earlygame.

### DIFF
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -294,7 +294,7 @@ butcher
 
 /datum/aiTask/sequence/goalbased/build/drone/precondition()
 	var/mob/living/critter/flock/F = holder.owner
-	return F?.can_afford(20)
+	return F?.can_afford(20) && (length(F?.flock?.units) > 10) //prioritise egg laying in the early game
 
 
 /datum/aiTask/sequence/goalbased/build/drone/get_targets()
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -294,7 +294,7 @@ butcher
 
 /datum/aiTask/sequence/goalbased/build/drone/precondition()
 	var/mob/living/critter/flock/F = holder.owner
-	return F?.can_afford(20) && F?.flock?.getComplexDroneCount() > 10) //prioritise egg laying in the early game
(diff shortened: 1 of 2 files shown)


code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -294,7 +294,7 @@ butcher
 
 /datum/aiTask/sequence/goalbased/build/drone/precondition()
 	var/mob/living/critter/flock/F = holder.owner
-	return F?.can_afford(20)
+	return F?.can_afford(20) && (length(F?.flock?.units) > 10) //prioritise egg laying in the early game
```
TobleroneSwordfish, line 297:
```suggestion
	return F?.can_afford(20) && F?.flock?.getComplexDroneCount() > 10) //prioritise egg laying in the early game
```

TobleroneSwordfish: slower, but probably worth it to be sure we're actually counting drones and not bits (could refactor `getComplexDroneCount` to be less stupid in future)

amylizzle: yeah, I might make `getComplexDroneCount()` less stupid in another PR, but sure I'll commit that.

code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -294,7 +294,7 @@ butcher
 
 /datum/aiTask/sequence/goalbased/build/drone/precondition()
 	var/mob/living/critter/flock/F = holder.owner
-	return F?.can_afford(20)
+	return F?.can_afford(20) && (F?.flock?.getComplexDroneCount() > 10) //prioritise egg laying in the early game
```
stonepillars, line 297: So the flock values egg more than secure room that spessmen can't break into to lay egg in?

amylizzle: yes, because the first 10 drones make it so much faster to convert and do flocky things.

TobleroneSwordfish: Also bear in mind you can tell a flockdrone to manually convert stuff or make flockbits to do it, but you can't tell a flockdrone to save up and lay an egg.

stonepillars (approved):

PR #300, opened 2022-05-19
Make complexDroneCount() not stupid. 

amylizzle:
## About the PR 
Turns flock.units into a associative list based on unit type - this lets us count drones or bits or w/e in O(1), instead of O(N), and maybe do other stuff later on. Also we can us `as anything` a lot more, which is good because it's cheaper.

## Why's this needed? 
Efficiency is important.

This is my last PR on this I swear.


ZeWaka (approved):

stonepillars (approved):
suitably jank. probably similar to `START_TRACKING_CAT` and related macros but whatever.

code/datums/flock/flock.dm
```diff
@@ -154,18 +156,19 @@
...
+			max_hp += F.max_health
 	if(max_hp != 0)
 		return hp/max_hp
 	else
 		return 0
 
 /datum/flock/proc/total_resources()
 	. = 0
-	for(var/mob/living/critter/flock/F as anything in src.units)
+	for(var/mob/living/critter/flock/F as anything in src.units[/mob/living/critter/flock/drone])
```
FlameArrow57, line 171:
```suggestion
	for(var/mob/living/critter/flock/drone/F as anything in src.units[/mob/living/critter/flock/drone])
```
Not necessary, but maybe since only expecting drones here?

code/datums/flock/flock.dm
```diff
@@ -409,14 +414,16 @@
 
 /datum/flock/proc/registerUnit(var/atom/movable/D)
 	if(isflock(D))
-		src.units |= D
+		if(!src.units[D.type])
+			src.units[D.type] = list()
```
FlameArrow57, line 418: Maybe could just initialize flockbits too in New(), since not expecting anything else than flockdrones and flockbits atm, or was it to just to prevent something like retyping every unit from isflock() in New()? Probably fine with either, guess it just didn't seem necessary to do a check on if type exists when there's only a few types available.

amylizzle: Not super needed, I only initialise the drone one because I directly reference it in code. If we need to reference bits also, we should do it then.

code/datums/flock/flock.dm
amylizzle: Yeah alright, but it'll need undoing if we ever add another flock critter that collects resources and isn't a drone

PR #301, opened 2022-05-19
Fixed some flockdrone absorber jank and nerfed paperbin health

TobleroneSwordfish:
[BUG]
## About the PR 
Fixes the flockdrone absorber not clearing its item reference on qdel, leading to it holding a qdeled item and causing jank. Since this was originally a paper bin bug I also stopped paper bins having 100 health because that seems a lot and I haven't learned my lesson about atomization.

## Why's this needed? 
infinitely absorbing paper bins getting stuck in drones is bad


amylizzle (approved):
Goddamnit, really? That's all that was missing?!

stonepillars (approved):

ZeWaka (approved):

Issue #140, opened 2022-04-10
Flock Control Panel Doesn't Clear Dead Enemies


amylizzle:
### Labels

[BUG]

### Description

If an enemy is dead/destroyed, it remains in the enemy list

### Steps to reproduce

1. Designate someone/something as enemy
2. Kill it
3. It's still an enemy
4. Destroy it's body
5. it's still an enemy and jumping to it causes a runtime

### Expected Behavior

Dead things should be removed

TobleroneSwordfish:
I'll wait until #123 is merged to try fix this, or it'll merge conflict horribly

FlameArrow57:
Should it remove dead enemies though? Not sure what Flock perspective is on human/etc. death, but if someone was to be revived, they wouldn't be marked as an enemy.

amylizzle:
> Should it remove dead enemies though? Not sure what Flock perspective is on human/etc. death, but if someone was to be cloned, they wouldn't be marked as an enemy.

That is a consideration, but some concessions should be made for the sake of usability imo. More specifically, the enemies panel ought to tell the player about active threats, not dormant ones. Also when the body is entirely destroyed it runtimes so that at least needs to be handled.

TobleroneSwordfish:
Currently I think drones will only attempt to cage a dead enemy if it's still marked as such, so it needs to persist through death. Could easily add a (dead) marker to the flock panel though.

amylizzle:
There's still a bug with this where you can jump to things that have a `null` loc and it causes a runtime.

amylizzle:
There's still a bug with this where you can jump to things that have a `null` loc and it causes a runtime.

PR #304, opened 2022-05-22
Fixes deleted enemies causing runtimes

TobleroneSwordfish:
[BUG]
## About the PR 
Some QDELETED checks that were missing and causing runtimes
Fixes what seems to have been determined as the actual issue from #140

## Why's this needed? 
runtimes bad

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -132,7 +132,7 @@
 	for(var/name in src.enemies)
 		var/list/enemy_stats = src.enemies[name]
 		var/atom/M = enemy_stats["mob"]
-		if(M && !QDELETED(M))
+		if(!QDELETED(M))
 			var/list/enemy = list()
 			enemy["name"] = M.name
 			enemy["area"] = enemy_stats["last_seen"]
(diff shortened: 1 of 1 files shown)


code/datums/flock/flock.dm
```diff
@@ -132,7 +132,7 @@
 	for(var/name in src.enemies)
 		var/list/enemy_stats = src.enemies[name]
 		var/atom/M = enemy_stats["mob"]
-		if(M)
+		if(M && !QDELETED(M))
```
stonepillars, line 135:
```suggestion
		if(!QDELETED(M))
```
QDELETED already null-checks.

code/datums/flock/flock.dm
```diff
@@ -65,7 +65,7 @@
 	switch(action)
 		if("jump_to")
 			var/atom/movable/origin = locate(params["origin"])
-			if(origin)
+			if(origin && !QDELETED(origin))
```
stonepillars, line 68:
```suggestion
			if(!QDELETED(origin))
```

stonepillars (approved):
though a little redundancy never hurt anyone

PR #305, opened 2022-05-22
Flock control panel now only sends relevent data

TobleroneSwordfish:
[CLEANLINESS]
## About the PR 
Makes it so the flock control panel only gets sent the data relevant to the current open tab, hopefully making it take a few less cycles.

## Why's this needed? 
The control panel is always going to be pretty heavy on server load so let's try to make it less so. Also stonepillars suggested it.


stonepillars:
!merge_upstream

stonepillars (approved):
lgtm.

stonepillars (commented):
Actually, why not also send the ui_tab in the data and use it to replace tabIndex? Since you call setTabIndex and change_tab action anyway?

stonepillars (approved):

PR #310, opened 2022-05-22
fix race condition with paper bins

amylizzle:
## About the PR 
Adds a check to make sure the item is on the ground before we try to eat it, in case someone already picked it up.

### DIFF
--- a/code/mob/living/critter/ai/flock/flocktasks.dm
+++ b/code/mob/living/critter/ai/flock/flocktasks.dm
@@ -659,6 +659,10 @@ stare
 
 /datum/aiTask/succeedable/harvest/on_tick()
 	var/obj/item/harvest_target = holder.target
+	if(!harvest_target || !isturf(harvest_target.loc)) //sometimes there are race conditions, and we should deal with that
+		harvest_target = null
+		fails++
+		return
 	if(harvest_target && BOUNDS_DIST(holder.owner, harvest_target) == 0 && !succeeded())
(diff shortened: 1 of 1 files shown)


code/mob/living/critter/ai/flock/flocktasks.dm
```diff
@@ -659,6 +659,10 @@ stare
 
 /datum/aiTask/succeedable/harvest/on_tick()
 	var/obj/item/harvest_target = holder.target
+	if(!harvest_target || !isturf(harvest_target.loc)) //sometimes there are race conditions, and we should deal with that
+		harvest_target = null
```
stonepillars, line 663: Redundant assignment to null right before a return?

amylizzle: not redundant, nulls out harvest_target to trigger failure() on the next evaluation and reset the task.

TobleroneSwordfish: surely harvest_target is reset anyway due to being a local variable?

stonepillars: You probably meant to null out `holder.target`, you mutated the value of `harvest_target`

amylizzle: oops, that is exactly what I meant to do. willfix.

amylizzle: oops, that is exactly what I meant to do. willfix.

PR #315, opened 2022-05-24
Refactors pathing and featherdoor access

TobleroneSwordfish:
[CLEANLINESS]
## About the PR 
Aloe pointed out the pathfinding code was wonky, this is the refactored version that also removes the `alien` var I added due to pathing jank.

## Why's this needed? 
f l o c k r e v i e w

### DIFF
--- a/code/obj/machinery/door/feather.dm
+++ b/code/obj/machinery/door/feather.dm
@@ -11,7 +11,6 @@
 	var/broken = FALSE
 	health = 80
 	health_max = 80
-	alien = TRUE
 
 /obj/machinery/door/feather/New()
 	..()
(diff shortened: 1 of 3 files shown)


code/obj/machinery/door/feather.dm
```diff
@@ -147,7 +149,9 @@
 ////////////////////
 // everyone allowed to open
 /obj/machinery/door/feather/friendly
```
FlameArrow57, line 151: Empty?

TobleroneSwordfish: empty?

FlameArrow57: It's just a proc with no body

FlameArrow57: Oh oops

frawhst (approved):

stonepillars (approved):

PR #92, opened 2022-03-23
Fix floorrunning bugs

FlameArrow57:
[BUG]

## About the PR 
This PR fixes several bugs with floorrunning.

Bugs fixed:
-Fixes being able to pick up stuff while floorrunning
-Fixes being able to throw stuff while floorrunning
-Fixes being able to interact with stuff like machinery while floorrunning
-Fixes being able to pull people while floorrunning
-Fixes people being able to pull you if you are floorrunning
-Fixes pinning not applying stuns for those involved when floorrunning is began
-Fixes being able to hit floorrunning drones with thrown objects
-Fixes being able to floorrun into tiles with dense walls, doors, windows, or steel grilles.
-Fixes being able to floorrun on broken Flock floor tiles and Flock walls
-Fixes breaking a tile with a Flockdrone floorrunning in it not ending floorrunning
-Fixes moving over and floorrunning through a tile with a Flockdrone floorrunning in it turning the tile off
-Fixes ending floorrunning on a floortile not turning the tile off if you didn't move off the tile

Also, to note what you can do right now:
-Floorrun into tiles with an object that has density, other than a wall, door, window, or steel grille
-Can carry items when floorrunning, but can't interact with them other than putting them in the disintegration reclaimer
-Can drop items when floorrunning
-Escape from pins with floorrunning

## Why's this needed? 
Bug fixes.

Fixes #58.
Fixes #59.
Fixes #85.

PR #319, opened 2022-05-24
Defines flockdrones as restrained while floorrunning

TobleroneSwordfish:
[BUG]
## About the PR 
Removes the cursed pull check in `atom.dm` and replaces it with defining drones as restrained while floorrunning.
This will prevent right click pulling too, if https://github.com/goonstation/goonstation/pull/8726 gets merged.

## Why's this needed? 
f l o c k r e v i e w

### DIFF
--- a/code/atom.dm
+++ b/code/atom.dm
@@ -584,11 +584,6 @@
 	if (iscarbon(usr) || issilicon(usr))
 		add_fingerprint(usr)
 
-	if (istype(usr, /mob/living/critter/flock/drone))
-		var/mob/living/critter/flock/drone/flockdrone = usr
-		if (flockdrone.floorrunning)
-			return
-
 	if (istype(src,/obj/item/old_grenade/light_gimmick))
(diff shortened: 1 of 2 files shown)


FlameArrow57:
Can you check if the click() implmented is needed?

Also if everything implemented in #92 still holds up?

TobleroneSwordfish:
Some of #92 doesn't actually seem to be working on master right now, let me poke it

TobleroneSwordfish:
Okay so it does seem to fulfil all of #92, and the click hack is still required to block throwing, although throwing probably shouldn't be allowed while restrained but that's another upstream issue.

FlameArrow57:
Cool - there is a can_throw var that mob critters have, could maybe use this instead of click stuff for now?

frawhst:
some critters can't throw things but I've always thought it was a bug. Lemme check

stonepillars (approved):

PR #321, opened 2022-05-24
Fixes some grab jank

TobleroneSwordfish:
[BUG]
## About the PR 
The flockdrone antigrab system wasn't actually working if the drone was AI controlled, since it was in client called movement code. I've moved it into the life loop, so now it just activates after a certain number of life loop calls.
Also fixed pins not working on flock tiles because it's related shut up.

## Why's this needed? 
no grabbing drones >:(

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -33,7 +33,7 @@
 
 	// antigrab powers
 	var/antigrab_counter = 0
-	var/antigrab_fires_at = 100
+	var/antigrab_fires_at = 3
 
 	var/glow_color = "#26ffe6a2"
 
@@ -399,6 +399,21 @@
 	if (src.dormant)
 		return
 
+	if(length(src.grabbed_by))
+		if (length(src.grabbed_by) == 1 && src.find_type_in_hand(/obj/item/grab/block))
+			src.antigrab_counter = 0
+		else
+			++src.antigrab_counter
+			if(src.antigrab_counter >= src.antigrab_fires_at)
+				playsound(src, "sound/effects/electric_shock.ogg", 40, 1, -3)
+				boutput(src, "<span class='flocksay'><b>\[SYSTEM: Anti-grapple countermeasures deployed.\]</b></span>")
+				for(var/obj/item/grab/G in src.grabbed_by)
+					var/mob/living/L = G.assailant
+					L.shock(src, 5000)
+				src.antigrab_counter = 0
+	else
+		src.antigrab_counter = 0
+
 	var/obj/item/I = absorber.item
 
 	if (!I)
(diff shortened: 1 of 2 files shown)


code/turf/turf_flock.dm
```diff
@@ -49,6 +49,10 @@
 	// do not call parent, this is not an ordinary floor
 	if(!C || !user)
 		return
+	if (istype(C, /obj/item/grab/))
```
FlameArrow57, line 52:
```suggestion
	if (istype(C, /obj/item/grab))
```

code/turf/turf_flock.dm
```diff
@@ -49,6 +49,10 @@
...
+		var/obj/item/grab/G = C
+		grab_smash(G, user)
```
FlameArrow57, line 54: Is this needed? Could just say grab_smash(C, user)?

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -386,6 +386,25 @@
...
+		src.antigrab_counter = 0
+		return
+
+	antigrab_counter++
+	if (antigrab_counter >= src.antigrab_fires_at)
+		playsound(src, "sound/effects/electric_shock.ogg", 40, 1, -3)
+		boutput(src, "<span class='flocksay'><b>\[SYSTEM: Anti-grapple countermeasures deployed.\]</b></span>")
+		for(var/obj/item/grab/G in src.grabbed_by)
+			if (istype(G, /obj/item/grab/block)) //do not shock ourselves
+				continue
```
FlameArrow57, line 402: I don't think you can have a block up if grabbed?

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -386,6 +386,25 @@
...
+			var/mob/living/L = G.assailant
+			L.shock(src, 5000)
+			qdel(G) //in case they don't fall over from our shock
```
FlameArrow57, line 405: Disorient doesn't drop the grab?

TobleroneSwordfish: not unless it knocks them over

code/mob/living/critter/flock/flockdrone.dm
TobleroneSwordfish: well then why are we checking length above :screm:

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -386,6 +386,23 @@
...
 
+///Runs the drone's inbuilt anti-grab measures, shocking the grabber after a while
+/mob/living/critter/flock/drone/proc/do_antigrab()
+	//if we're blocking that means we're not grabbed
+	if (!length(src.grabbed_by) || src.find_type_in_hand(/obj/item/grab/block))
+		src.antigrab_counter = 0
+		return
+
+	antigrab_counter++
+	if (antigrab_counter >= src.antigrab_fires_at)
```
FlameArrow57, line 397:
```suggestion
	src.antigrab_counter++
	if (src.antigrab_counter >= src.antigrab_fires_at)
```
Very minor, not necessary, but was kind of using some interchanging

TobleroneSwordfish: shame on me smh

code/mob/living/critter/flock/flockdrone.dm
```diff
@@ -399,6 +416,8 @@
 	if (src.dormant)
 		return
 
+	src.do_antigrab()
```
stonepillars, line 419: This proc is only used here, consider in-lining it? It just adds proc call overhead.

stonepillars, line 419: This proc is only used here, consider in-lining it? It just adds proc call overhead.

stonepillars (approved):

PR #324, opened 2022-05-25
Azrun Code Review - Achievements

amylizzle:
## About the PR 
Makes achievements be defines, and alters a comment to better reflect updated code

### DIFF
--- a/code/datums/flock/flockunlockable.dm
+++ b/code/datums/flock/flockunlockable.dm
@@ -2,8 +2,9 @@ ABSTRACT_TYPE(/datum/unlockable_flock_structure)
 /**
  * Subclass this for every new building type you add
  * Override the check_unlocked() function to do whatever unlock logic you have
- * If you're looking for a specific event, I recommend /datum/flock.achievements
+ * If you're looking for a specific event, I recommend /datum/flock.hasAchieved()
  */
+
 /datum/unlockable_flock_structure
 	var/structType = null
 	var/datum/flock/my_flock = null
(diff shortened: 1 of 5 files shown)


code/datums/flock/flockunlockable.dm
```diff
@@ -52,11 +53,11 @@ ABSTRACT_TYPE(/datum/unlockable_flock_structure)
...
 	check_unlocked()
-		return src.my_flock.hasAchieved("all_structures")
+		return src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CHEAT_STRUCTURES)
 
 /datum/unlockable_flock_structure/gnesisturret
 	structType = /obj/flock_structure/gnesisturret
 
 	check_unlocked()
-		return src.my_flock.hasAchieved("human dissection") || src.my_flock.hasAchieved("all_structures")
+		return src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CAGE_HUMAN) || src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CHEAT_STRUCTURES)
```
stonepillars, line 62:
```dm
return ..() || src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CAGE_HUMAN)
```
can't something like this be done and then move `src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CHEAT_STRUCTUERS)` to the parent?

amylizzle: I guess - is there any point though? seems like it'd just add more proc overhead and not really fix the issue of needing to copy & paste when you want to define a new structure unlock

stonepillars: It feels a little cleaner and I have feeling this is preferred by imcoder to copy+pasting conditions everywhere. Probably best to ask there. If they say otherwise my review's moot and it fine to be wordy.

code/datums/flock/flockunlockable.dm
```diff
@@ -52,11 +53,11 @@ ABSTRACT_TYPE(/datum/unlockable_flock_structure)
 	structType = /obj/flock_structure/compute
 
 	check_unlocked()
-		return src.my_flock.hasAchieved("all_structures")
+		return src.my_flock.hasAchieved(FLOCK_ACHIEVEMENT_CHEAT_STRUCTURES)
```
stonepillars, line 56: Identical to parent implementation now, you can remove it.

ZeWaka (approved):

stonepillars (approved):

PR #333, opened 2022-05-28
Aloe Review 2

amylizzle:
## About the PR 
Changes gnesis turret syringes to barbed, and slight buffs regen rate because oh my god was it slow
Removes old //TODO
Changes cage to use proper reagent init


stonepillars (approved):

ZeWaka (approved):
morb

PR #334, opened 2022-05-28
Revert nerfs to flockdrone absorption

TobleroneSwordfish:
[BALANCE]
## About the PR 
Partially reverts the nerfs to flockdrone resource absorption from #270.
Flockdrones now gain 4 resources per health absorbed.
The flock rift now comes with an extra egg containing 3 flockbits.

## Why's this needed? 
Flockdrones were replicating at an extremely slow rate, meaning they were basically unable to sustain any real losses incurred. 
The nerf cut the rate at which flock could gain resources by around 4 times, this caused flock to not really work anymore.

### DIFF
--- a/code/mob/living/critter/flock/flockdrone.dm
+++ b/code/mob/living/critter/flock/flockdrone.dm
@@ -26,7 +26,8 @@
 	butcherable = TRUE
 
 	var/health_absorb_rate = 2 // how much item health is removed per tick when absorbing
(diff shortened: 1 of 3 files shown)


amylizzle:
I am very strongly against the absorbtion bonus, for reasons of the behaviour it encourages (splitting stacks for max efficiency)

Also plz tweak the egg to spit out some flockbits, maybe at the cost of a drone.

stonepillars (approved):
plays funner me likey

PR #336, opened 2022-05-29
Allow flock to work as a random event antag

TobleroneSwordfish:
[BUG]
## About the PR 
Enables the flockmind random event antag spawn (for admin spawns only) and changes some stuff to make it work.

## Why's this needed? 
Flockmind admin spawns are a good way of testing flock without devoting a whole round to the (currently probably very imbalanced) flock gamemode

### DIFF
--- a/code/datums/flock/flock.dm
+++ b/code/datums/flock/flock.dm
@@ -4,6 +4,8 @@
 
 /// associative list of flock names to their flock
 /var/list/flocks = list()
+/// Has a flock relay been unleashed yet this round
+var/signal_unleashed = FALSE
 
 /// manages and holds information for a flock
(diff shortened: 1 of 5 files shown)


code/datums/flock/flock.dm
```diff
@@ -4,6 +4,8 @@
 
 /// associative list of flock names to their flock
 /var/list/flocks = list()
+/// Has a flock relay been unleashed yet this round
+var/signal_unleashed = FALSE
```
ZeWaka, line 8: if you're making a new global at least prefix it with flock

amylizzle (approved):
oh no, globals

stonepillars (approved):

PR #337, opened 2022-05-29
Added attack feedback to flock structures

TobleroneSwordfish:
[FEATURE]
## About the PR 
Adds a shake animation when flock structures are hit with something and makes them use a crunchy glass sound when they are attacked with something heavy.

## Why's this needed? 
Attack feedback good, makes it obvious that this is a thing you can and should destroy.

### DIFF
--- a/code/obj/flock/cage.dm
+++ b/code/obj/flock/cage.dm
@@ -12,6 +12,7 @@
 	health_max = 30
 	alpha = 192
 	anchored = FALSE
+	hitTwitch = FALSE
 	var/atom/occupant = null
 	var/obj/target = null
 	var/eating_occupant = FALSE
--- a/code/obj/flock/structure/flock_structure_parent.dm
+++ b/code/obj/flock/structure/flock_structure_parent.dm
@@ -202,7 +202,12 @@ ABSTRACT_TYPE(/obj/flock_structure)
 		damtype = "fire"
 
 	takeDamage(damtype, W.force)
-	playsound(src.loc, "sound/impact_sounds/Crystal_Hit_1.ogg", 50, 1)
+	hit_twitch(src)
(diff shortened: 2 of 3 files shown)


FlameArrow57:
Idk, do they need twitch feedback? I don't think there's anything else that really has twitch feedback, plus they have hit sound, attack particle, and chat message already

TobleroneSwordfish:
It was feedback we got from the playtest, and players not knowing what they should be hitting is a longstanding problem with flock iirc

amylizzle (approved):

ZeWaka (approved):

FlameArrow57:
Okay. Idk if this the right place to put the hit_twitch(), as cages already have a twitch thing on takeDamage, so hit twitch stuff would occur twice for them

stonepillars (approved):

code/obj/flock/cage.dm
```diff
@@ -12,6 +12,7 @@
 	health_max = 30
 	alpha = 192
 	anchored = FALSE
+	hitTwitch = FALSE
```
FlameArrow57, line 15: I think a clarifying comment saying that it implements its own twitch stuff would be nice

code/obj/flock/structure/flock_structure_parent.dm
```diff
@@ -17,6 +17,8 @@ ABSTRACT_TYPE(/obj/flock_structure)
 	var/health = 30
 	var/health_max = 30
 	var/bruteVuln = 1.2
+	///Should it twitch on being hit?
```
FlameArrow57, line 20:
```suggestion
	///Should it twitch on being hit through attackby()?
```
Or something similar?

PR #338, opened 2022-05-29
Cage + Turret Cost Balance

amylizzle:
## About the PR 
Doors now have 200 health (up from 80) - ~30ish toolbox hits
Turrets cost 150 each
Cages inject their occupant with gnesis when they get down to ripping out flesh parts
Cages are harder to break out of now - you will definitely lose a sizeable portion of items if you get caged

## Why's this needed? 
Balance

### DIFF
--- a/code/obj/flock/cage.dm
+++ b/code/obj/flock/cage.dm
@@ -195,6 +195,7 @@
 			else if(occupant && ishuman(occupant))
 				var/mob/living/carbon/human/H = occupant
 				getHumanPiece(H) //cut off a human part and add it to contents, set it to target
+				H?.reagents?.add_reagent(target_fluid,2) //you get a bit of juice, just to complicate life
 			else if(occupant && isrobot(occupant))
 				var/mob/living/silicon/robot/H = occupant
 				getRobotPiece(H) //cut off a robot part and add it to contents, set it to target
@@ -268,8 +269,9 @@
 	relaymove(mob/user as mob)
 		if (user.stat)
 			return
-
-		if(prob(25))
+		if(ON_COOLDOWN(src,"move_damage",1 SECONDS))
+			return
+		if(prob(75))
 			takeDamage("brute",1)
 		return
 
--- a/code/obj/machinery/door/feather.dm
+++ b/code/obj/machinery/door/feather.dm
@@ -9,8 +9,8 @@
 	mat_changedesc = FALSE
 	autoclose = TRUE
 	var/broken = FALSE
-	health = 80
-	health_max = 80
+	health = 200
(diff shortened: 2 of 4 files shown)


code/obj/machinery/door/feather.dm
```diff
@@ -9,8 +9,8 @@
 	mat_changedesc = FALSE
 	autoclose = TRUE
 	var/broken = FALSE
-	health = 80
-	health_max = 80
+	health = 200
+	health_max = 200 //half of a standard door
```
TobleroneSwordfish, line 13:
```suggestion
	health_max = 200 //a third of a standard door
```
airlocks have 600 health

code/obj/flock/cage.dm
```diff
@@ -268,8 +269,9 @@
 	relaymove(mob/user as mob)
 		if (user.stat)
 			return
-
-		if(prob(25))
+		if(ON_COOLDOWN(src,"move_damage",1 SECONDS))
```
FlameArrow57, line 272:
```suggestion
		if(ON_COOLDOWN(src,"move_damage",1 SECOND))
```

code/obj/machinery/door/feather.dm
FlameArrow57: Honestly idk if comment is needed at all, or just could say fraction of door health, seems like an implementation detail thing

FlameArrow57: Honestly idk if comment needed at all, or could just say fraction of door health, seems like an implementation detail thing using specific fraction

code/obj/flock/cage.dm
```diff
@@ -268,8 +269,9 @@
...
+			return
+		if(prob(75))
 			takeDamage("brute",1)
```
FlameArrow57, line 275:
```suggestion
		if(!ON_COOLDOWN(src,"move_damage",1 SECOND))
			if(prob(75))
				takeDamage("brute",1)
```
SECONDS -> SECONDS. Also, otherwise ignores first attack I think?

code/obj/flock/cage.dm
FlameArrow57, line 272:
```suggestion
		if(ON_COOLDOWN(src,"move_damage",1 SECOND))
```

code/obj/machinery/door/feather.dm
stonepillars: People can hover to see the typical value of door health.

amylizzle: I'll just remove the comment

stonepillars (approved):

ZeWaka (approved):

PR #339, opened 2022-05-29
Anti-birdwatching balance

TobleroneSwordfish:
[BALANCE]
## About the PR 
Prevents AIs from tracking flockmobs and makes converting a tile break any cameras on that tile.

## Why's this needed? 
I played AI during one of the playtests and realised just how vulnerable flock would be to an AI reporting their location constantly whenever they start trying to set up. This doesn't actually change balance much, since flockdrones could already just pick up some wirecutters, but this just makes it so any flock converted area is automatically obscured from prying AI eyes.
Also having 20 "floating gewgaw" in the tracking list isn't very helpful to anyone.

can move this upstream if the main PR gets merged before it/pillars tells me to


code/mob/living/silicon/ai.dm
```diff
@@ -2315,6 +2315,8 @@ proc/get_mobs_trackable_by_AI()
 			continue
 		if (M == usr)
 			continue
+		if (isflockmob(M))
```
stonepillars, line 2318: Apply `PROP_MOB_AI_UNTRACKABLE` to flockthings instead.

code/mob/living/intangible/flockmob_parent.dm
```diff
@@ -23,6 +23,7 @@
 	src.blend_mode = BLEND_ADD
 	REMOVE_ATOM_PROPERTY(src, PROP_MOB_INVISIBILITY, src)
 	APPLY_ATOM_PROPERTY(src, PROP_MOB_INVISIBILITY, src, INVIS_FLOCK)
+	APPLY_ATOM_PROPERTY(src, PROP_MOB_AI_UNTRACKABLE, src)
```
stonepillars, line 26: Aren't all intangible things untrackable?

